PR #4, opened 2017-10-25
Feature/token overlay

Velenir:
Refactoring of TokenOverlay component to make it more modular.

### DIFF
--- a/src/components/TokenList/index.tsx
+++ b/src/components/TokenList/index.tsx
@@ -1,0 +1,18 @@
+import React from 'react'
+import TokenItem from '../TokenItem'
+import { code2tokenMap, TokenCode } from 'globals'
+
+
+interface TokenListProps {
+  tokens: TokenCode[],
+  balances: {[code in TokenCode]: number }
+}
+
+const TokenList: React.SFC<TokenListProps> = ({ tokens, balances }) => (
+  <div className="tokenList">
+    {tokens.map(code =>
+      <TokenItem name={code2tokenMap[code]} code={code} balance={balances[code]} key={code} />)}
+  </div>
+)
+
+export default TokenList
--- a/src/components/TokenOverlay/index.tsx
+++ b/src/components/TokenOverlay/index.tsx
@@ -1,21 +1,23 @@
 import React, { Component } from 'react'
-import TokenItem from '../TokenItem'
+import TokenOverlayHeader from '../TokenOverlayHeader'
+import TokenList from '../TokenList'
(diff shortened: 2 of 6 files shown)


src/styles/components/home/_home.scss
```diff
@@ -421,7 +421,7 @@
       position: fixed;
     }
 
-    > ul {
+    > .tokenList {
```
W3stside, line 424: You verified this won't conflict somewhere else? Mainly concerned as this is indented CSS styling

src/components/TokenList/index.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import TokenItem from '../TokenItem'
+import { code2tokenMap, TokenCode } from 'globals'
+
+
+interface TokenListProps {
+  tokens: TokenCode[],
+  balances: {[code in TokenCode]: number }
```
W3stside, line 8: this is TS typing for an Object within an Object yeah?

src/components/TokenOverlay/index.tsx
```diff
@@ -1,29 +1,30 @@
...
-  GNT: 'GOLEM',
-}
+import TokenOverlayHeader from '../TokenOverlayHeader'
+import TokenList from '../TokenList'
+import { code2tokenMap, TokenCode } from 'globals'
+import { createSelector } from 'reselect'
+
+const filterTokens = createSelector(
+  (state: TokenOverlayState, _: TokenOverlayProps) => state.filter.toUpperCase(),
+  (_: TokenOverlayState, props: TokenOverlayProps) => props.tokenCodeList,
```
W3stside, line 9: what is the '_' used as param 1?

Velenir: Just a convention meaning `_` param isn't going to be used. The first two arrow function must have the same signature as that is what `filterTokens` will receive as input and pass onto them.

src/components/TokenList/index.tsx
Velenir: balances is an object with keys, that are any of TokenCode values, that is:
```
balances = {
  ETH: 6578,
  GNO: 12345,
  ...
}
```

it is kinda like array comprehension in python

src/styles/components/home/_home.scss
Velenir: So far as token list isn't used anywhere else, it won't conflict

W3stside (approved):
Yay - looks solid, works well. Good stuff

PR #5, opened 2017-10-25
Feature/wallet - separation of Header Components

W3stside:
Separated Header into 4 components
1. Header wrapper - renders the other 3
2. MenuWallet - shows Account and Balance
3. MenuAuctions - shows current auctions and tokens
4. Hamburger - hamburger button + sliding panel


src/components/Hamburger/index.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { Component } from 'react'
+
+interface HamburgerProps {}
+
+interface HamburgerState {
+  isOpen: boolean
+}
+
+export default class Hamburger extends Component<HamburgerProps, HamburgerState> {
+  constructor(props: HamburgerProps) {
```
Velenir: It is mostly a matter of taste, but a shorter alternative would be to use class props:
```
state = { isOpen: false}
handleClick = () => this.State({ isOpen: !this.state.isOpen })
```
and no need for constructor

src/components/MenuWallet/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  balance: Object,
+  tokens: Object,
+}
+
+export const MenuWallet: React.SFC<WalletProps> = ({ account, balance, tokens }) => (
+  <i className="menuWallet">
+    {/* Wallet Info - Address && Balance */}
+    <span>
+      <code>{`${account ? account.slice(0,14) : 'loading...'}...`}</code>
+      <small>{balance ? balance : 'loading...'} ETH</small>
```
Velenir: it'll display `loading...` when `balance === 0`

stories/MenuAuction.tsx
```diff
@@ -0,0 +1,53 @@
...
+  </div>
+)
+
+const constructKnobs = (
+  name = 'YOUR AUCTIONS',
+  auctionsArr: Array<any>,
+  separator: any
+) => ({
+  name: text('title', name),
+  ongoingAuctions: array('Ongoing Auctions', auctionsArr, separator)
```
Velenir: `ongoingAuctions` knob displays `[Object]` as it JSON.stringifies it like that
storybook-knobs don't allow to edit array of objects easily
a workaround would be to add a knob for each auction:
```
ongoingAuctions: auctionsArr.map((auction, i) => object(`Ongoing Auction ${i}`, auction)),
```

also would be nice to have more auctions, even randomly generated

Velenir (approved):

PR #7, opened 2017-10-26
Feature/button cta

Velenir:
ButtonCTA component with a story
The anchor tag inside accepts all props passed to the component for easy config

### DIFF
--- a/src/components/ButtonCTA/index.tsx
+++ b/src/components/ButtonCTA/index.tsx
@@ -1,0 +1,37 @@
+import React, { PureComponent } from 'react'
+
+interface ButtonCTAProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+  children: React.ReactNode,
+  className?: string,
+  onClick(e: React.SyntheticEvent<HTMLAnchorElement>): void
+}
+
+class ButtonCTA extends PureComponent<ButtonCTAProps> {
+  static defaultProps = {
+    className: 'blue',
+  }
+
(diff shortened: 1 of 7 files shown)


src/components/ButtonCTA/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+    onClick && onClick(e)
+  }
+
+  render() {
+    const { className, children, onClick, ...rest } = this.props
+
+    return (
+      <a
+        href="#"
+        className={'buttonCTA ' + (className || '')}
```
W3stside, line 27: Without defaultProp could be ``className={`buttonCTA${(className ? ` ${className}` : '')}`}`` to avoid extra whitespace issue
##### OR otherwise no conditional is necessary since there is defaultProp no? 
template styling - again just stylistic
```className={`buttonCTA ${className}`}```

W3stside (approved):
Noice

src/components/ButtonCTA/index.tsx
Velenir: While it is a microoptimization, having static `defaultProps` is marginally faster than creating values on demand.
See more at [SO](https://stackoverflow.com/a/42733457/7316062)

In case you wanted to have `<a class="buttonCTA">` without `blue`, you would pass a falsy className: 
e.g. `<ButtonCTA className=""/>` or `<ButtonCTA className={null}/>`, right?
But `buttonCTA ${null}` ` === "buttonCTA null"`, so I chose `className || ''`

PR #8, opened 2017-10-26
Feature/TopAuctions

Velenir:
TopAuctions component

All token pairs will be received from redux store, sorted inside `mapStateToProps` and then passed to the component

### DIFF
--- a/stories/TopAuctions.tsx
+++ b/stories/TopAuctions.tsx
@@ -1,0 +1,59 @@
+import * as React from 'react'
+
+import { storiesOf, StoryDecorator } from '@storybook/react'
+import { object } from '@storybook/addon-knobs'
+
+import TopAuctions from 'components/TopAuctions'
+
+import { code2tokenMap } from 'globals'
+
+/*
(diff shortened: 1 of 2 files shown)


stories/TopAuctions.tsx
```diff
@@ -0,0 +1,59 @@
...
+  if (code === 'ETH') return acc
+
+  acc[`${code}/ETH`] = +Math.random().toFixed(8)
+  return acc
+}, {}) as { [pair: string]: number }
+
+
+/**
+ * sorts pairs by ASC and takes top 5
+ * @param {typeof ratioPairs} pairs
```
W3stside, line 27: lol we really should have been documenting this way before as well - or wait I guess that's  what TS is for 😂

W3stside (approved):
Looks good!

stories/TopAuctions.tsx
Velenir: Yep, I just added it after a half hour break I came back and didn't immediately understand what my code was doing

PR #9, opened 2017-10-27
Home Page Text + No Wallet Components

W3stside:
Created components/NoWallet + components/TextSquare + stories/[both]

### DIFF
--- a/src/components/NoWallet/index.tsx
+++ b/src/components/NoWallet/index.tsx
@@ -1,0 +1,23 @@
+import React from 'react'
+
+import ButtonCTA from 'components/ButtonCTA'
+
+interface NoWalletProps {
+  handleClick?(): void
+  hide?: boolean
+}
+
+export const NoWallet: React.SFC<NoWalletProps> = ({ 
+  handleClick,
+  hide = false, 
+}): JSX.Element => (
+  <div className={`noWallet${hide ? ' hide' : ''}`}>
+    <h2>Enable &amp; unlock your <strong>Mist / Metamask wallet</strong> to connect before continuing.</h2>
+    <span className="icon-walletUnlock"></span>
+    <ButtonCTA onClick={handleClick}>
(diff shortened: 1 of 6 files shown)


stories/TextSquare.tsx
```diff
@@ -0,0 +1,46 @@
...
+    }}
+  >
+    {story()}
+  </div>
+)
+
+storiesOf('TextSquare', module)
+  .addDecorator(CenterDecor)
+  .addWithJSX('TextSquareLeft[Intro]', () => 
+    <section className="home">
```
Velenir: `<section>` can be added once as a decorator, before `CenterDecor`

src/components/NoWallet/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+interface NoWalletProps {
+  handleClick?(): void
+  hide?: boolean
+}
+
+export const NoWallet: React.SFC<NoWalletProps> = ({ 
+  handleClick,
+  hide = false, 
+}): JSX.Element => (
```
Velenir: `JSX.Element` is unnecessary as `React.SFC` provides everything we need, including return values

Velenir (approved):

stories/TextSquare.tsx
Velenir: But then you'll have to do
```
.addWithJSX('TextSquareBoth', (): any =>
    [<TextSquare />, <NoWallet handleClick={action('ButtonCTA clicked')} hide={boolean('Hide/Show', false)} />],
)
```
So it's your call

W3stside: Thanks - had it as a decorator before but changed it back instead of going the [] approach. Will most likely change to keep DRY - cheers

src/components/NoWallet/index.tsx
W3stside: cool thanks

PR #10, opened 2017-10-27
Added MockStore @ stories/helpers/mockStore

W3stside:
1. added `initialState` 2nd parameter to `store.js` function to add initial state values
2. created `stories/helpers/mockStore.ts` for adding initialState values and importing
3. created shorter store creation function

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -9,7 +9,7 @@ import Notifications from 'middlewares/Notifications'
 
 import reducer from 'reducers'
 
-export default function (history) {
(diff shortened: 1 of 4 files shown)


Velenir (approved):
All's great, just change `const req = require.context('../stories', true, /\.tsx$/)` to use `false` in `.storybook/config.js`. So we don't pull in from subdirs (`helpers`). Granted, right now there are no `.tsx` files to match, but if we move decorators there'll be.

PR #18, opened 2017-11-01
Feature/connected TopAuctions

Velenir:
1. Changed RatioPairs structure
2. Made TopAuctions onClick quick-select token pair from an auction


src/reducers/tokenPair.ts
```diff
@@ -1,18 +1,19 @@
 import { handleActions } from 'redux-actions'
 
-import { selectTokenAndCloseOverlay } from 'actions'
+import { selectTokenAndCloseOverlay, selectTokenPair } from 'actions'
 import { TokenPair } from 'types'
 import { TokenItemProps } from 'components/TokenItem'
 
-export default handleActions<TokenPair, TokenItemProps>(
+export default handleActions<TokenPair, TokenItemProps & TokenPair>(
```
W3stside, line 7: normally these .toString() are NOT required - I know you were running into issues with reducers but are you sure this is necessary? Normally just inputting the action export name is enough...

Velenir: It is either `.toString()` or `as any`. And `.toString()` is called either way, when casting object keys.

I don't mind changing if you feel it would be better

W3stside (approved):
Merge!

src/reducers/tokenPair.ts
W3stside: in terms of micro op is `.toString()` faster?

PR #19, opened 2017-11-01
Feature/home container

W3stside:
committed similar messages twice - second commit should read:
`edited Home component, container + story`

### DIFF
--- a/src/api/initialization.ts
+++ b/src/api/initialization.ts
@@ -36,7 +36,7 @@ const contractArtifacts = [
 ].map(name => require(`../../build/contracts/${name}.json`))
 
 class DutchExchangeInit {
-  contracts: Object | any
+  contracts: object | any
   web3: any
   constructor() {
     this.contracts = _.fromPairs(contractArtifacts.map((artifact) => {
@@ -64,6 +64,7 @@ class DutchExchangeInit {
   async fireUp(opts?: Object | any) {
     await this.setWeb3Provider(opts.ethereum)
   }
+
   /** ASYNC Class Method: setWeb3Provider
    * @param provider => comes from Redux <opts.ethereum>
    * @param {any} provider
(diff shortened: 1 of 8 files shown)


src/api/initialization.ts
```diff
@@ -36,7 +36,7 @@ const contractArtifacts = [
 ].map(name => require(`../../build/contracts/${name}.json`))
 
 class DutchExchangeInit {
-  contracts: Object | any
+  contracts: object | any
```
Velenir, line 39: `object | any === any`
Not much point in it. Feel free to add something like 
```
interface Contracts{ DutchExchange: any, DutchExchangeFactory: any }
```
to types

Velenir (approved):

src/selectors/blockchain.ts
```diff
@@ -22,6 +22,8 @@ export const findDefaultProvider = (state: State) => {
   })
 }
 
+export const getActiveProvider = (state: State): Providers[string] => selector(state).activeProvider
```
Velenir: `Providers[string]` is `any`
To grab from Providers values you need `Providers[keyof Providers]`
Another way would be to specify possible providers:
```
type ProviderName = 'METAMASK' | 'MIST'
interface Providers {
 [P in ProviderName]: any
}
```
then `(state: State): ProviderName`
also we'll need Provider interface at some point

Velenir (approved):

src/api/initialization.ts
W3stside: yeah there're probably alot of these kinds of ts typings laying around from older code ill hunt through these commit specific ones

src/selectors/blockchain.ts
W3stside: achso ok ill take a look and edit these, cheers

PR #22, opened 2017-11-02
Feature/change token state

W3stside:
1. changed State shape to reflect removal of Tokens type to use TokenBalances outside of Blockchain interface
2. reworked components/containers to include above changes
3. removed unnecessary conditional rendering logic in `components/MenuWallet` due to reducer `initialState`


Velenir (approved):
Looks good! Merge

PR #23, opened 2017-11-03
Component/closing price bar

W3stside:


src/components/ClosingPriceBar/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export interface ClosingPriceBarProps {
+  sellToken: TokenCode,
+  sellTokenPrice: Balance,
+  buyToken: TokenCode,
+}
+
+const ClosingPriceBar: React.SFC<ClosingPriceBarProps> = ({ sellToken, sellTokenPrice, buyToken }) => 
+  <div className="auctionLastPrice">
+    <small>Closing price of last auction:</small>
```
Velenir: On `PANEL 3` we have this same component but with `<small>`~~Closing~~ `price of last auction:</small>`
It would be nice to vary the text based on some kind of prop or pass it in as `children`

Velenir (approved):

src/components/ClosingPriceBar/index.tsx
W3stside: you know what, good call I saw your implementation of the `panel variation` in your Auction `story` and I'm gonna go with that

PR #26, opened 2017-11-03
Feature/auction status

Velenir:
AuctionStatus presentational component + story

### DIFF
--- a/src/components/AuctionStatus/index.tsx
+++ b/src/components/AuctionStatus/index.tsx
@@ -1,0 +1,53 @@
+import React from 'react'
+
+import { TokenCode, Balance } from 'types'
+
+import claim from 'assets/claim.svg'
+
+// TODO: move to global types
+export enum Status {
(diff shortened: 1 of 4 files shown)


src/components/AuctionStatus/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+    </span>
+
+    <span>
+      <small>STATUS</small>
+      <big data-status={status}>{capitalize(status)}</big>
+    </span>
+
+    <span>
+      {
+        status === Status.ACTIVE ? [
```
W3stside: May be fine here as the conditional choices are rather limited but I prefer to create ENUM components and split up the components
ex. ```const AuctionStatus = ({status}) => {
  [status.ACTIVE]: <AuctionActive>,
  [status.XYZ]: <AuctionXYZ>
}```

W3stside (approved):
Merge!

PR #29, opened 2017-11-03
Feature/auction selling getting

W3stside:
thoughts? 
1. internal state for AuctionSellingGetting

### DIFF
--- a/src/components/AuctionSellingGetting/index.tsx
+++ b/src/components/AuctionSellingGetting/index.tsx
@@ -1,0 +1,68 @@
+import React, { Component } from 'react'
+
+import { Balance, TokenCode } from 'types'
+
+/* CONSIDER ADDING GAS_COST */
+export interface AuctionSellingGettingProps {
+  balance: Balance,
+  buyToken: TokenCode,
+  ratio: number,
+  sellToken: TokenCode,
+}
+
+export interface AuctionSellingGettingState {
+  value: string | number
+}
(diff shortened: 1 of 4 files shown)


src/components/AuctionSellingGetting/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const { value } = this.state
+
+    return (
+      <div className="auctionAmounts">
+        <span>
+          <label htmlFor="sellingAmount">Amount Selling:<a href="#max" onClick={this.onClick}>MAX</a></label>
+          <input 
+            type="number" 
+            name="sellingAmount" 
+            id="sellingAmount"
```
Velenir: I'd really like to be able to pass `<input ... min="0" max={balance}/>`, to restrict values in a native way.
But the problem is, this leads to `input` being resized based on `max` value.
Also another problem  -- on smaller screens absolutely positioned element (especially `<small>{sellToken}</small>`) are getting overlapped by static elements.

I have a proposal to significantly change this element's template and styles.

One way is to still use absolute positioning, but leave an unshrinkable gap using `padding`:
https://jsfiddle.net/Velenir/m4tu78zv/

Another way is to use CSS Grids and make every relevant element into `.auctionAmount`'s direct child, which will be resized dynamically :
https://jsfiddle.net/Velenir/ozLvzs7x/

What do you think?

PR #34, opened 2017-11-10
Test/karma setup

Velenir:
Karma setup and auction test reworked to work in browser with **Metamask** and without

Steps

Without **Metamask**:

1. start *testrpc* deterministically, e.g. `testrpc -a 3 -d`
2. `truffle compile --all` to recompile contracts just in case
3. `truffle migrate`
4. run test without **Metamask** `npm run test:karma`

> between tests it is generally enough to call `truffle migrate --reset`, but a repeat of 1-3 steps may be needed in some cases

-------------------

With **Metamask**:
1. start **testrpc** deterministically, e.g. `testrpc -a 3 -d`
2. `truffle compile --all` to recompile contracts just in case
3. `truffle migrate`
4. import accounts to **Metamask**
4. `npm run test:karma:manual`
5. open Chrome to `http://localhost:9876/debug.html`
6. open console and follow instructions. You need to have a particular account active when it is the *from* account in the transaction object. Otherwise you'll have `Unknown address - unable to sign transaction for this address`. **Metamask** can't handle multiple accounts at once.

You may encounter `<contractName> is not deployed` error, then going through 1-3 steps may help.
Also in **Metamask** some transactions may error with `account nonce and tx nonce are different` or some such, then you need to go through steps 1-3 and change network in **Metamask** from `localhost:8545` to a different one and then back.

> between tests it is generally enough to call `truffle migrate --reset` and refresh the page, but a repeat of 1-3 steps may be needed in some cases


W3stside (approved):
Good

PR #36, opened 2017-11-14
Test/karma tests

Velenir:
Auction creation, submitting orders, completion and claiming funds tests.

Works with testrpc and with some problems with Metamask.

### DIFF
--- a/src/test/browser/ETH2GNO.test.ts
+++ b/src/test/browser/ETH2GNO.test.ts
@@ -88,6 +88,12 @@ describe('ETH 2 GNO contract', () => {
     gno = await GNO.deployed()
     tul = await TUL.deployed()
 
+    Object.assign(accs, { dx: DX.address, eth: ETH.address, gno: GNO.address, tul: TUL.addresss })
+
+    watchAllEventsFor(dx, 'DutchExchange')
+    watchAllEventsFor(eth, 'ETH')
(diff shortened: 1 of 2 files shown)


src/test/browser/ETH2GNO.test.ts
```diff
@@ -489,4 +495,26 @@ describe('ETH 2 GNO contract', () => {
     console.log(`= ${totalETH}\t${totalGNO}`)
   }
 
+  function watchAllEventsFor(contract: any, name: string) {
+    const addr2acc = Object.entries(accs).reduce((accum, [name, addr]) => (accum[addr] = name, accum), {})
```
W3stside: Not sure how you feel about this but let me know either way, would love to see an example of what these reduce fns spit out, sometimes when going through reviews its a bit tricky here

W3stside (approved):
Merge

src/test/browser/ETH2GNO.test.ts
Velenir: Given `accs = {master: '0x23df...'}`  it flips `key/val` transforming into `addr2acc = { '0x23df...': "master"}`
So that we can substitute hard to read addresses with account names in event watchers

W3stside: `addr 2 acc` makes sense 😅

PR #38, opened 2017-11-15
Feature/contract token balance

W3stside:
1. on load actions fire to grab token balances and dispatch to state
2. reducer created to add each token balance

TODO:
1. Proper typing on some fns
2. refactoring some functions

### DIFF
--- a/src/api/dutchx.ts
+++ b/src/api/dutchx.ts
@@ -1,7 +1,7 @@
 import dutchX from './initialization'
 import { weiToEth } from 'utils/helpers'
 
-let dutchXInst: any
+let dxInst: any
 
 /**
  * Initializes connection to DutchX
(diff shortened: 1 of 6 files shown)


src/api/dutchx.ts
```diff
@@ -10,9 +10,9 @@ let dutchXInst: any
 export const initDutchXConnection = async (DUTCHX_OPTIONS: any) => {
   console.log(' ===> FIRING initDutchX ACTION')
   try {
-    dutchXInst = await dutchX.init(DUTCHX_OPTIONS)
-    return dutchXInst
-    // console.log('SUCCESS CONNECTING TO DUTCHX INSTANCE', dutchXInst)
+    dxInst = await dutchX.init(DUTCHX_OPTIONS)
```
Velenir: Actually, as we are returning `dxInst` from an async function, it's the same as returning `Promise.resolve(dxInst)`
So, no need for all that logic:
```
let promisedDX: Promise<any> 

export const initDutchXConnection = async (DUTCHX_OPTIONS: any) => {
   console.log(' ===> FIRING initDutchX ACTION')
   try {
    return promisedDX = dutchX.init(DUTCHX_OPTIONS)
  }
  ....

export { promisedDX }
```

Then there'll be no need for `getDutchXConnection`, we'll just need to `await promisedDX`
Or you can keep `export const getDutchXConnection = () => promisedDX`, if you think a function looks better semantically, no need for `async`, though.

Velenir (approved):

src/api/dutchx.ts
W3stside: OK fair enough - removed async from the `getDutchXConnection` and will remove some unnecessary comments/console logs

W3stside:
@Velenir reverted and changed 2 things, looks good

PR #39, opened 2017-11-16
Feature/connected panels

Velenir:
Connected components to routes
Now can click through, but without interaction with tokens
Don't reload page on subroutes, need to implement redirects where necessary first

### DIFF
--- a/src/components/AuctionPanel/index.tsx
+++ b/src/components/AuctionPanel/index.tsx
@@ -12,7 +12,7 @@ interface AuctionPanelProps {
 
 const AuctionPanel: React.SFC<AuctionPanelProps> = ({ auctionAddress }) => (
   <AuctionContainer auctionDataScreen="status">
-    <AuctionHeader backTo="/">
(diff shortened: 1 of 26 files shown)


src/components/AuctionPanel/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import AuctionProgress from 'containers/AuctionProgress'
+import AuctionStatus from 'containers/AuctionStatus'
+
+interface AuctionPanelProps {
+  auctionAddress: string
+}
+
+const AuctionPanel: React.SFC<AuctionPanelProps> = ({ auctionAddress }) => (
+  <AuctionContainer auctionDataScreen="status">
+    <AuctionHeader backTo="/">
```
W3stside: should be previous page `/order`

Velenir: Yep, my misstake

W3stside (approved):
Good, merge

PR #40, opened 2017-11-17
Feature/auction amount summary

Velenir:
Changes:

1. Added `AuctionAmountSummary` component that looks like a static `TokenPair` with already set selling and buying amounts.
2. `TokenPair` state now includes `sellAmount` field, which together with closing price makes up estimated `buyAmount`
2. Made `AuctionSellingGetting` into a controlled component, so that we can pull `sellAmount` whenever we need it, e.g. in `AuctionAmountSummary`, and later when submitting sell order.
3. Added `TokenOverlay` to `PANEL 2`
4. misc fixes

### DIFF
--- a/src/components/OrderPanel/index.tsx
+++ b/src/components/OrderPanel/index.tsx
@@ -19,7 +19,7 @@ const OrderPanel: React.SFC<OrderPanelProps> = ({ sellToken, buyToken }) => (
   <AuctionContainer auctionDataScreen="amount">
     <TokenOverlay />
     <AuctionHeader backTo="/">
-      Token Auction ${sellToken}/${buyToken}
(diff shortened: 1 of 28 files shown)


src/components/OrderPanel/index.tsx
```diff
@@ -16,6 +17,7 @@ interface OrderPanelProps {
 
 const OrderPanel: React.SFC<OrderPanelProps> = ({ sellToken, buyToken }) => (
   <AuctionContainer auctionDataScreen="amount">
+    <TokenOverlay />
     <AuctionHeader backTo="/">
       Token Auction ${sellToken}/${buyToken}
```
W3stside: templating issue, you forgot the templating quotes

W3stside (approved):

PR #45, opened 2017-11-22
Code changes

hacker-DOM:
Change overflow, handle case when no sell orders, reset buyerBalances and claimedAmounts, and getUnclaimedTokens functionality


contracts/DutchExchange.sol
```diff
@@ -145,11 +136,14 @@ contract DutchExchange {
         public 
     {
         postBuyOrder(amount, _auctionIndex);
-        claimBuyerFunds(_auctionIndex);
+        claimBuyerFunds(msg.sender, _auctionIndex);
     }
 
-    function claimSellerFunds(uint256 _auctionIndex) public returns (uint256 returned) {
-        uint256 sellerBalance = sellerBalances[_auctionIndex][msg.sender];
+    function claimSellerFunds(address user, uint256 _auctionIndex)
```
W3stside, line 142: why the change for `user` parameter? Won't this technically allow anyone to `claimSellerFunds`?

hacker-DOM: yes, but we're sending it to `user` - whoever owns the funds
so you can give anyone their own funds, but you will pay for the gas
(it has to be done like this because M&S wanted the top-level claimAllFunds etc. functions to be able to take in any user)

W3stside:
@dteiml  Can this be closed and deleted?

josojo:
old...

PR #50, opened 2017-11-26
Feature/truffle scripts2

Velenir:
Added a script for watching events

+ `truffle exec trufflescripts/watch_events.js` on all contracts
+ `... --dx` on a specific contract
+ `... --dx NewSellOrder,AuctionCleared` for specific events on a specific contract

> **Note on behavior**
> it starts watching from the *latest* block, so if there are already events in that block, they'll be printed immediately on script execution

### DIFF
--- a/trufflescripts/watch_events.js
+++ b/trufflescripts/watch_events.js
@@ -19,7 +19,6 @@ const argv = require('minimist')(process.argv.slice(2), { alias: { v: 'verbose' 
  */
 
 module.exports = async () => {
-  console.log(argv)
   // web3 is available in the global context
   const [master, seller, buyer] = web3.eth.accounts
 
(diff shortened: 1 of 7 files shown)


trufflescripts/watch_events.js
```diff
@@ -0,0 +1,155 @@
...
+      } else event.watch(printLog.bind(null, name))
+    }
+
+    const eventsArray = typeof events === 'string' ? events.split(',') : ['allEvents']
+
+    eventsArray.forEach((e) => {
+      try {
+        processEvent(contract[e](filterObj))
+      } catch (error) {
+        if (error.message === 'contract[e] is not a function') {
```
W3stside: haven't tested but should be `if (error.message === '${contract[e]} is not a function') {...}` - no?

Velenir: No, that's the actual message, even if it isn't helpful at all. Stack trace works in mysterious ways.

PR #52, opened 2017-11-27
Refactor/testrpc snapshotting

W3stside:
1. Revert auction by: `revert_auction.js`
2. `start_auction.js` will automatically snapshot


trufflescripts/utils/index.js
```diff
@@ -26,9 +26,18 @@ module.exports = (web3) => {
     increaseTimeBy(increaseBy)
   }
 
+  const makeSnapshot = () => web3.currentProvider.send({ jsonrpc: '2.0', method: 'evm_snapshot' }).result
+
+  const revertSnapshot = async (blockN = '0x01') => {
```
Velenir: Not a block number
`send` is synchronous, no need to `await`
if you specifically want async, use `sendAsync`

trufflescripts/start_auction.js
```diff
@@ -19,6 +19,8 @@ module.exports = async () => {
 
   // auctionStart is in the future
   if (timeUntilStart > 0) {
+    const snapshot = await makeSnapshot()
+    console.log(`SNAPSHOT CREATED OF BLOCK # ${snapshot}`)
```
Velenir: I don't think snapshot corresponds to block #
It's just an incrementing id.
`makeSnapshot` is sync, no need to await, same with `revertSnapshot` above

trufflescripts/revert.js
```diff
@@ -0,0 +1,38 @@
...
+ * -a seller|buyer|<address>    for the given address
+ * -i <index>                   for auction with given index
+ * --last                       for last auction
+ */
+
+module.exports = async () => {
+  let blockID
+
+  blockID = argv.b
+  if(argv.b === undefined) blockID = '0x01'
```
Velenir: Can be written as `const blockID = argv.b || '0x01'`
or `const blockID = argv.b !== undefined ? argv.b || '0x01'` if you want to check for undefined

Velenir (approved):
Looks good, however unreliable snapshotting is

PR #53, opened 2017-11-30
Refactor/dutch x

Velenir:
WIP

more abstraction and modularity for:
+ web3
+ Tokens
+ DutchX
+ an overarching contracts/account api in `api/index.js`


src/api/DutchX.ts
```diff
@@ -34,7 +34,7 @@ async function init(): Promise<DutchExchange> {
   }
 
   const getClosingPrice = async (pair: TokenPair, index?: number) => {
-    if (index === undefined) index = await getAuctionIndex(pair)
+    if (index === undefined) index = await getAuctionIndex(pair) - 1
```
W3stside: this is what needs to be taken out yeah?

src/api/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  // account would normally be taken from redux state and passed inside an action
+  // but just in case
+  if (!account) account = await web3.getCurrentAccount()
+
+  return Tokens.getTokenBalance('ETH', account)
+}
+
+// TODO: remove ['ETH', 'GNO'] default, use actions for this
+export const getTokenBalances = async (codeList: TokenCode[] = ['ETH', 'GNO'], account?: Account) => {
```
W3stside: why is this called codeList and not tokenList?

Velenir: don't know, will change to `tokenList`

src/api/DutchX.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    return index.toNumber()
+  }
+
+  const getClosingPrice = async (pair: TokenPair, index?: number) => {
+    if (index === undefined) index = await getAuctionIndex(pair) - 1
+
+    const [num, den] = getExchange(pair).closingPrices(index)
+
+    return num.div(den).toString()
```
W3stside: Not super positive with the changes that will come - but I think we should be careful of this return of `num/den` as we may need to divide the inverse when calculating the other side of Token Exchanges _e.g `ETH/GNO` && `GNO/ETH`_

src/api/DutchX.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    return vol.toString()
+  }
+
+  const getSellerBalances = async (pair: TokenPair, index?: number, account?: Account) => {
+    [index, account] = await fillDefaultIndexAndAccount(pair, index, account)
+
+    const balance = await getExchange(pair).sellerBalances(index, account)
+
+    return await balance.toString()
```
W3stside: why `await` here?

src/api/web3Provider.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const resetProvider = () => setProvider(getProvider())
+
+  return {
+    getCurrentAccount,
+    getAccounts,
+    getETHBalance,
+    getNetwork,
+    isConnected,
+    get currentProvider() {
```
W3stside, line 56: This is new to me - are you declaring the function here _and_ calling the getter?

Velenir: declaring the getter, same as
```
class CL {
  get prop () {
    return this.prop
  }
}
```

src/api/DutchX.ts
Velenir: nice spotting skills. will remove

src/api/contracts.ts
```diff
@@ -0,0 +1,38 @@
...
+  'DutchExchange',
+  'DutchExchangeFactory',
+  'DutchExchangeETHGNO',
+  'DutchExchangeGNOETH',
+  'Token',
+  'TokenETH',
+  'TokenGNO',
+]
+
+const Contracts = contractNames.map(name => TruffleContract(require(`../../build/contracts/${name}.json`)))
```
W3stside: Could you rename to `ContractJSON` or `ContractArtifacts` ? 😅

Velenir: These are not contract artifacts, they are contract classes, if you will. Since each is `TruffleContract(artifact)`

W3stside: oh I didn't see the TruffleContract before, sneaky. Nvm then

src/api/utils/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  window.addEventListener('load', function loadHandler(event) {
+    window.removeEventListener('load', loadHandler, false)
+    return accept(event)
+  }, false)
+})
+
+
+export const promisify = (func: Function, context: object, ...defArgs: any[]) =>
```
W3stside, line 17: I hate this function 😂 😂 😂 😂

Velenir: I'm open to any changes. This one is actually a slightly modified `util.promisify` from node, but we can't use node system libraries in the browser(

src/components/BalanceButton/index.tsx
```diff
@@ -1,6 +1,6 @@
 /* eslint no-console:0 */
```
W3stside: can you delete this component, not used

W3stside: can you delete this component, not used

src/api/utils/index.ts
W3stside: its fine for now

W3stside (approved):
From testing and going through code it looks good, merge

PR #68, opened 2017-12-18
extrapolated some overused logic into outside scope functions

W3stside:
1. moved repetitive logic into functions

### If this is MORE confusing and convoluted please tell me and I'll change it back

### DIFF
--- a/test/dutchExchange.js
+++ b/test/dutchExchange.js
@@ -16,43 +16,92 @@ const { wait } = require('@digix/tempo')(web3)
 const MaxRoundingError = 100
 
 // Test VARS
-let sellToken
-let buyToken
+let eth
+let gno
 let dx
 let oracle
 
+// testing Auction Functions
+const setupTest = async (accounts) => {
+  // get buyers, sellers set up and running
+  gno = await TokenGNO.deployed()
+  eth = await EtherToken.deployed()
+  // create dx
+  dx = await DutchExchange.deployed()
+  // create price Oracle
+  oracle = await PriceOracle.deployed()
+
+  for (let acct = 1; acct < accounts.length; acct++) {
+    await Promise.all([
+      gno.transfer(accounts[acct], 10 ** 18, { from: accounts[0] }),
+      // depoit into etherToken contract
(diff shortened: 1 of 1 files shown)


test/dutchExchange.js
```diff
@@ -16,43 +16,92 @@ const { wait } = require('@digix/tempo')(web3)
...
+  // create price Oracle
+  oracle = await PriceOracle.deployed()
+
+  for (let acct = 1; acct < accounts.length; acct++) {
+    await Promise.all([
+      gno.transfer(accounts[acct], 10 ** 18, { from: accounts[0] }),
+      // depoit into etherToken contract
+      eth.deposit({ from: accounts[acct], value: 10 ** 9 }),
+      // depositing into the exchange
+      eth.approve(dx.address, 10 ** 9, { from: accounts[acct] }),
```
Velenir, line 40: Strictly speaking `dx.deposit(eth` should go after `eth.approve(dx`, because it depends on having `allowance[acc][dx]` in the token. It works here for you because *testrpc* handles these promises sequentially.
But it would be a good idea to `eth.approve().then(...dx.deposit()` or something.

Velenir (approved):

test/dutchExchange.js
josojo: For the deposit into the etherToken contract, we do not need any approval, since here no erc20 tokens are send, just ether

josojo (approved):
thanks a lot for the cleanings.

test/dutchExchange.js
Velenir: not etherToken, just dx. It's how github displays lines previous to the comment that may be confusing.
What I mean to change is:
```
eth.approve(dx.address, 10 ** 9, { from: accounts[acct] }).then(() => dx.deposit(eth.address, 10 ** 9, { from: accounts[acct] })),
gno.approve(dx.address, 10 ** 18, { from: accounts[acct] }).then(() => dx.deposit(gno.address, 10 ** 18, { from: accounts[acct] })),
```

josojo: Got your point! Sure, we need that! Since I merged this pull request already, I will push it in my new commit

PR #69, opened 2017-12-18
minor fixes

Velenir:
fixes for minor issues here and there

### DIFF
--- a/migrations/2_deploy_contracts.js
+++ b/migrations/2_deploy_contracts.js
@@ -23,7 +23,7 @@ module.exports = function deploy(deployer, networks, accounts) {
     .then(() => PriceOracle.deployed())
     .then((p) => {
       PriceOracleInstance = p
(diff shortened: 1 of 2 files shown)


migrations/2_deploy_contracts.js
```diff
@@ -23,7 +23,7 @@ module.exports = function deploy(deployer, networks, accounts) {
     .then(() => PriceOracle.deployed())
     .then((p) => {
       PriceOracleInstance = p
-      return deployer.deploy(OWL, TokenGNO.address /* ,PriceOracle.adress */)
+      return deployer.deploy(OWL, TokenGNO.address /* ,PriceOracle.adress */, 0)
```
Velenir, line 26: if we don't provide the expected number of arguments to a solidity function, it won't just default the missing arg to its initial value (0 here) but all arguments (both GNOaddress and Oracle address = 0)

josojo (approved):
good catches! Thanks

PR #81, opened 2017-12-21
Contracts/feature/new solidity code from dom

josojo:
try do make the migration run and do the merge together


contracts/DutchExchange.sol
```diff
@@ -507,41 +494,39 @@ contract DutchExchange {
         public
         returns (uint returned, uint tulipsIssued)
     {
-        // Requirements
         uint sellerBalance = sellerBalances[sellToken][buyToken][auctionIndex][user];
+        // R1
         require(sellerBalance > 0);
 
         // Get closing price for said auction
         fraction memory closingPrice = closingPrices[sellToken][buyToken][auctionIndex];
```
Velenir: I think `closingPrice` here can be declared with `storage` instead of `memory`.
`memory` creates a copy of the closing price in contract's storage; `storage` -- only a reference.

Same on lines 633, 674, 786, 787

contracts/DutchExchange.sol
```diff
@@ -320,51 +303,58 @@ contract DutchExchange {
...
+            require(auctionIndex == latestAuctionIndex);
         } else {
-            // check whether current auchtion has no starting date, because auctions have not been funded.
-            if (auctionIndex == latestAuctionIndices[orderedPairFristElement(buyToken, sellToken)][orderedPairSecondElement(buyToken, sellToken)]){
-                  require ( auctionStarts[orderedPairFristElement(buyToken, sellToken)][orderedPairSecondElement(buyToken, sellToken)] == 1);
+
+
+            // C2
+            // R2.1: sell orders must go to next auction
+            if (getAuctionStart(sellToken, buyToken) < now && closingPriceDen > 0 && closingPriceDenOpp > 0){
```
Velenir: Currently `getAuctionStart(sellToken, buyToken) < now` is always true at least for the first, freshly created toke pair, so I can repeatedly post sell orders to the ongoing auction.

contracts/DutchExchange.sol
```diff
@@ -634,125 +628,87 @@ contract DutchExchange {
     )
         public
         constant
-        returns (uint num, uint den)
+        returns (fraction price)
```
Velenir: With these changes functions that return `structs` (`getPrice` and `priceOracle`) can't be called from JavaScript:

```
invalid solidity type!: tuple
```

`(uint num, uint den)`, which was returned previously, is properly converted into `BigNumber[2]`
It would be nice to see either return to the `(uint, uint)` convention or a helper function - wrapper over `getPrice`, something like:
```
function getPriceForJS(...) public constant returns (uint, uint) {
    fraction memory price = getPrice(...);
    return (price.num, price.den);
}
```

contracts/DutchExchange.sol
```diff
@@ -764,166 +720,200 @@ contract DutchExchange {
...
     }
-    
+
     /// @dev Gives best estimate for market price of a token in ETH of any price oracle on the Ethereum network
     /// @param token address of ERC-20 token
     /// @return Weighted average of closing prices of opposite Token-ETH auctions, based on their sellVolume  
     function priceOracle(
         address token
+    )
+        public
```
Velenir: This function is no longer `constant`, so can't be called from JavaScript to get a meaningful result, only transaction receipt. Is this change intentional? Also it returns a `struct`, same issue as above with `getPrice` function.

contracts/DutchExchange.sol
```diff
@@ -113,39 +108,41 @@ contract DutchExchange {
...
     event NewTokenPair(
         address indexed sellToken,
         address indexed buyToken
     );
-    event AuctionCleared(address indexed sellToken, address indexed buyToken, uint indexed auctionIndex);
+
+    event AuctionCleared(
+        address sellToken,
+        address buyToken,
+        uint sellVoluem,
```
Velenir: typo sellVolu~em~me

contracts/DutchExchange.sol
josojo: ran into the same issues. Will write a wrapper function, since we need the returns fraction, to save variables.

contracts/DutchExchange.sol
josojo: We used memory because it stores the struct only temporarily and is cheaper to use than storage

contracts/DutchExchange.sol
josojo: Got help from stefan and rimeissner:
this should be possible with the new solidity version
0.4.19
https://github.com/ethereum/solidity/blob/develop/Changelog.md#0419-2017-11-30

contracts/DutchExchange.sol
josojo: solved this one in a seperate commit

contracts/DutchExchange.sol
josojo: solved in newest commit

contracts/DutchExchange.sol
josojo: made it constant again

PR #105, opened 2018-01-05
refactored contract functions out into sep file

W3stside:
Refactored out contract code into sep file

### DIFF
--- a/test/testFunctions.js
+++ b/test/testFunctions.js
@@ -17,7 +17,7 @@ const contractNames = [
  * deploys all contracts in build/contracts
  */
 const getContracts = async () => {
-  const depContracts = contractNames.map(c => artifacts.require(c)).map(async cc => cc.deployed())
+  const depContracts = contractNames.map(c => artifacts.require(c)).map(cc => cc.deployed())
   const contractInstances = await Promise.all(depContracts)
 
   const deployedContracts = contractNames.reduce((acc, name, i) => {
--- a/test/testFunctions.js
+++ b/test/testFunctions.js
@@ -1,0 +1,171 @@
+/* eslint prefer-const:0, max-len:0, object-curly-newline:1, no-param-reassign:0, no-console:0 */
+const { wait } = require('@digix/tempo')(web3)
+const { timestamp } = require('./utils')
+
+const MaxRoundingError = 100000
+
+const contractNames = [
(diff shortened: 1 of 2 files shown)


josojo:
I think the fact that the third test command is not going through has nothing to do with your specific implementation. Had these issues before. Maybe accounts ran out of funding. Idk.
  
I like the refactoring a LOT

test/testFunctions.js
```diff
@@ -0,0 +1,171 @@
...
+  'TokenTUL',
+  'PriceOracle',
+]
+
+/**
+ * >getContract()
+ * deploys all contracts in build/contracts
+ */
+const getContracts = async () => {
+  const depContracts = contractNames.map(c => artifacts.require(c)).map(async cc => cc.deployed())
```
Velenir: `async` is unnecessary here

test/testFunctions.js
```diff
@@ -0,0 +1,171 @@
...
+const setupTest = async (accounts, { DutchExchange: dx, EtherToken: eth, TokenGNO: gno, PriceOracle: oracle }) => {
+  // Await ALL Promises for each account setup
+  await Promise.all(accounts.map((acct) => {
+    /* eslint array-callback-return:0 */
+    if (acct === accounts[0]) return
+
+    eth.deposit({ from: acct, value: 10 ** 9 })
+    eth.approve(dx.address, 10 ** 9, { from: acct })
+    gno.transfer(acct, 10 ** 18, { from: accounts[0] })
+    gno.approve(dx.address, 10 ** 18, { from: acct })
```
Velenir, line 47: `await Promise.all` on line 38 doesn't ultimately wait for the `eth.* gno.*` transactions, but only starts them. `accounts.map` maps to `Array<undefined>` not promises. Thanks to the sequential nature of `testrpc`'s transaction handling, it won't usually break anything, but the right thing would be to wrap them in `Promise.all()`:

```
return Promise.all([eth.deposit, eth.approve, gno.transfer, gno.approve])
```

Same with lines 51, 52 below.

test/testFunctions.js
W3stside: yep no idea why thats in there O.O

test/testFunctions.js
W3stside: true true

PR #107, opened 2018-01-09
Merge refactored code into "master" branch

hacker-DOM:


hacker-DOM:
I will merge because we need the merge to diagnose the bug :/

PR #111, opened 2018-01-09
small fix, opposite auctions are closed in case auctionstart==1

josojo:
just a one liner change that I need for my tests


josojo:
Dom did not pulled this, but coded it quickly by himself.

PR #112, opened 2018-01-09
Tests/fuzzy/assert rejects

Velenir:
+ Added recognition of transactions with `Log` event as rejected
+ Fixed linting in fuzzy tests (mainly hoisted functions to the beginning of file)
+ Simplified `if-else` logic
+ Added `evm_mine` instruction after `evm_increaseTime` as timestamp isn't actually increased otherwise

-----------------
+ no changes to test logic

Currently fail up to 8-11 fuzzy tests
To tests only fuzzy tests: `truffle test test/fuzzy.js`
To test without fuzzy tests: `truffle test test/dutchExchange.js test/dutchExchange-Tulip.js`


josojo (approved):
I checked all changes. Looks very good. 

although I can not judge 100% about the fuzzy.js refactoring, since I do not know the code well. But from what I have seen, it looks good

josojo:
@Velenir you can go ahead an merge

Issue #120, opened 2018-01-10
mint Tulips after auction has cleared


hacker-DOM:
saves gas

josojo:
agree with it - let's change this.

W3stside:
@dteiml @josojo closing as done

Issue #119, opened 2018-01-10
Users won't be able to claim extraTokens


hacker-DOM:
We currently require(returned > 0) in claimBuyerFunds (R1). But extraTokens can only be computed after auction has cleared. If a user claims all of his to-date balance & the auction clears *in the same block*, he/she won't have access to extraTokens

josojo:
right

Issue #116, opened 2018-01-10
not all extraTokens will be reclaimable


hacker-DOM:
Arbitrating opposite auction results in an increase in buyVolume
We use a trader's proportion of buyVolume to calculate his extraTokens
As a result, not all extraTokens will be claimable

hacker-DOM:
not relevant in latest specs

PR #122, opened 2018-01-10
Solved issues #116, #119, #120

hacker-DOM:


W3stside:
@dteiml That's ok man, thanks for opening this, will test it shortly

contracts/DutchExchange.sol
```diff
@@ -463,34 +464,47 @@ contract DutchExchange {
 
         // We have to compute the price at intersection time,
         // which will be exactly half of initial price
-        fraction memory sellTokenPrice = priceOracle(sellToken);
-        fraction memory buyTokenPrice = priceOracle(buyToken);
-        fraction memory price;
-
-        price.num = sellTokenPrice.num * buyTokenPrice.den;
-        price.den = sellTokenPrice.den * buyTokenPrice.num;
+        fraction memory price = computeRatioOfHistoricalPriceOracles(sellToken, buyToken, auctionIndex);
```
josojo, line 467: Can you please explain? the oustandingVolume is calculated with the getPrice of the current time. Then we can not calculate the outstandingVolumeOpp with a completely new price, right?

josojo: Hm thought about it again. Probably we should calculate both oustandingVolume and outstandingVOlumeOpp with the priceOracle.

hacker-DOM: That's exactly what this does ;) I just reduced code duplication by extracting it into its own function (since getPrice uses the same mechanism, it can just call this fn).

josojo: I think you did this only for the outstandingVolumeOpp, but not for the outstandingVolume. This one is still calculated within the postBuyOrder with getPrice.

josojo: It's quite confusing to so many different price function, which all call another pricefunction. Though I am not sure, whether we can reduce complexity

hacker-DOM: I'm with you on the last one, but solving bugs with the smallest possible complexity is what I do 9 to 5 :D 
I'll think about the first comm, if it should be different

josojo:
These commits are not logner relevant after new specification discussed on 1.11

PR #128, opened 2018-01-11
Silent test run

Velenir:
Tor suppress logging, run
```
truffle test --silent
```
or
```
truffle test -s
```

Also please use `log` from `./uitls` instead of `console.log` in tests.


josojo:
Silent option works for me only with testrpc, but not with ganach-cli

besides this, all good!

josojo:
silent modus in truffle develop can be added later

PR #129, opened 2018-01-11
Contracts/test/new spec fixing tests

josojo:
-- Integrates the first TradeFlow tests from the excel file: Flows 1,6,9

-- removes the test/dutchExchange.js, since it will eventually be replaced by the TradeFlow tests

-- tried to clean the tests from the dutchExchange.js as much as possible.


josojo:
found the error, all good

contracts/DutchExchange.sol
```diff
@@ -579,32 +512,32 @@ contract DutchExchange {
     {
         (returned, tulipsToIssue) = getUnclaimedBuyerFunds(sellToken, buyToken, user, auctionIndex);
 
-        // R1
-        // require(returned > 0);
-        if (returned == 0) {
-            Log('claimBuyerFunds R1');
-            return;
-        }
+        uint den = closingPrices[sellToken][buyToken][auctionIndex];
```
Velenir: should be
```
uint den = closingPrices[sellToken][buyToken][auctionIndex].den;
```

contracts/DutchExchange.sol
```diff
@@ -817,7 +748,7 @@ contract DutchExchange {
         // (Takes in a amount of user's TUL tokens as ration of all TUL tokens, outputs fee ratio)
         // We premultiply by amount to get fee:
         if (totalTUL > 0) {
-            feeRatio.num = totalTUL - 10 * balanceOfTUL;
+            feeRatio.num = Math.atleastZero(int(totalTUL - 10 * balanceOfTUL));
             feeRatio.den = 16000 * balanceOfTUL + 200 * totalTUL;
```
Velenir: `uint balanceOfTUL = TokenTUL(TUL).lockedTULBalances(user);` from a couple lines above can be moved inside `if (totalTUL > 0) {`

Would save a function call in case of `else`

contracts/DutchExchange.sol
```diff
@@ -206,7 +204,7 @@ contract DutchExchange {
             fraction memory priceToken2 = priceOracle(token2);
 
             // Compute funded value in ETH and USD
-            uint fundedValueETH = token1Funding * priceToken1.num / priceToken1.den + token2Funding * priceToken2.num / priceToken2.den;
+            uint fundedValueETH = token1Funding * priceToken1.num / priceToken1.den; + token2Funding * priceToken2.num / priceToken2.den;
```
Velenir: extra `;`

contracts/DutchExchange.sol
hacker-DOM: yes

contracts/DutchExchange.sol
hacker-DOM: yyes

contracts/DutchExchange.sol
hacker-DOM: yup

josojo: lol, it only throws a warning, but no compilation error.

josojo:
its all in david's branch tulip testing

PR #131, opened 2018-01-12
Contracts/tests/token aprroval

Velenir:
Tests of token approval (`DutchExchange.updateApprovalOfToken`)

Run with `truffle test test/dutchExchange-TokenApproval.js -s`


josojo (approved):
Very clean work! like it. All tests cases are covered.

PR #134, opened 2018-01-13
enforcing invariant: 

josojo:
balances in the exchange without deposits/withdraws on smart contract level are adding up to a constant

->currently js makes small errors since numbers are too big. Will switch to BigNum.

### DIFF
--- a/contracts/DutchExchange.sol
+++ b/contracts/DutchExchange.sol
@@ -673,16 +673,17 @@ contract DutchExchange {
 
         // if (opposite is zero auction OR opposite auction has cleared) {
         if ((sellVolumesCurrent[buyToken][sellToken] == 0 && opp.den == 0) || opp.den > 0) {
+            
+            if (sellVolumesCurrent[buyToken][sellToken] == 0 && opp.den == 0) {
+                extraTokens[buyToken][sellToken][auctionIndex + 1] += extraTokens[buyToken][sellToken][auctionIndex];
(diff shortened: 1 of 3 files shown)


W3stside:
@josojo does my addition of `Number.prototype.toWei` or `toEth` in PR #137 help this at all?

josojo:
I think only bigInt will help here. unfortunately.

contracts/DutchExchange.sol
```diff
@@ -680,15 +673,18 @@ contract DutchExchange {
 
         // if (opposite is zero auction OR opposite auction has cleared) {
         if ((sellVolumesCurrent[buyToken][sellToken] == 0 && opp.den == 0) || opp.den > 0) {
```
hacker-DOM: I don't think we need opp.den == 0?

hacker-DOM: other than that, loooks good.

hacker-DOM: in fact, it isn't even logically correct. If opp.den > 0, the second condition will fire ...

josojo: I agree, we do not need it. I added it, because it increases readability:
(sellVolumesCurrent[buyToken][sellToken] == 0 && opp.den == 0)==> dealing with 0 sell order
opp.den > 0 ==>opposite auction was already closed.

josojo: anyway, I am okay with taking && opp.den out.

PR #142, opened 2018-01-15
Contracts/test/fix bugs

hacker-DOM:


josojo (approved):
Yes, exactly what we need.

PR #146, opened 2018-01-16
Contracts/test/fix bugs

josojo:
DOm will push his latest code into this PR.

### DIFF
--- a/contracts/DutchExchange.sol
+++ b/contracts/DutchExchange.sol
@@ -508,9 +508,9 @@ contract DutchExchange {
         uint auctionIndex
     )
         public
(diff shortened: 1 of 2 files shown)


josojo:
https://github.com/gnosis/dutch-exchange/blob/62f51f2055058fc6ecfd09fc3aa041d475c201d1/contracts/DutchExchange.sol#L858

there is something wrong with the new price function. 

the starting price of an auction would be 
'Price at this moment === ' 1.6
instead of 2 right at the beginning of the auction

josojo:
also please take out the terms, which are canceling themselves in nom and den

PR #151, opened 2018-01-17
Contracts/feature/maker oracle integration

josojo:
this is the new Oracle implementation integrating the MakerDAO oracle Pricefeed.

Design decisions:

-We have an interface contract talking to the MakerContract. Advantage: if the makerDAO decides to replace their infrastructure, we can just change our interfaceContract.

-PriceOracle delivers just the rounded ETH-USD price, cents are not considered, since they are neglectable. 

- feeInUSD is calculated with 18 decimal places, as OWL has also 18 decimal places. So that is a good fit.


josojo:
Tradeflows are passing,
depositWithdraws are tested

josojo:
its all in davids code

PR #160, opened 2018-01-19
Contracts/tests/tulip testing refactor

W3stside:
### tulip testing refactor
1. added withdrawUnlockedTUL tokens testing
- [x] - Passes?
2. removed getLockedTul getter from TokenTUL.sol in favour of mapping var call


josojo:
How do you know that tests are not throwing a Log event indicating reverts?

You need this one, right?

    eventWatcher(dx, 'Log', {})
  after(eventWatcher.stopWatching)

It's also missing at nearly all my tests :(

let's talk about this in the standup, maybe we will just switch to reverts instead of logging....


I like that you also tested the split of tulips in case you have two buyers:
    assert.equal((b1TulipsIssued + b2TulipsIssued).toEth(), 99.5, 'Tulips only issued / minted after auction Close so here = 99.5 || sell Volume')

Should this also be check this split for two sellers? Not sure whether this isn't implicitly tested through.

Like the way you tested it with all these transparent logs.

W3stside:
@josojo had to change to `afterEach()` as my `after()` hook was failing. Will update the others if necessary

W3stside:
@josojo I'd like to switch the `require` back into the conditional `if` blocks in a sep PR, can I merge this if you've tested it?

W3stside:
@josojo ok the test is done - contract 5 run via `truffle test test/dutchExchange-Tulip.js -c 5 -s`

PR #165, opened 2018-01-23
Contracts/helper func for contract flag

Velenir:
Dynamically running contracts (`truffle test file.js -c1`) can be enabled with:
```
const { enableContractFlag } = require('./utils')

const c1 = () => contract('...', () => {...})
...

enableContractFlag(c1, c2, ...)  // order matters
```


W3stside (approved):
nice, thanks

PR #166, opened 2018-01-24
Contracts/test/get price

josojo:
just added the tests for getPrice functions.

PR #167, opened 2018-01-24
Contracts/test/price oracle

josojo:
tests the function historicalPriceOracle.

priceOracle is not needed to test since it is basically only calling historicalPriceOracle


Velenir:
Does this PR include code from Contracts/test/get price #166 PR?

josojo:
resubmitted with new smart contracts

PR #170, opened 2018-01-25
Contracts/tests/post sell order

Velenir:
Tests for `postSellOrder` function
Bonus: flag `--noevents` to skip printing event logs


josojo (approved):
very clean work

there are some overlaps with stateTransition tests, but I think this is a good thing.

PR #171, opened 2018-01-25
Contracts/test/claim buyer funds test

josojo:
just review pull request and unfortunately all the tests form priceOracle and historicalPriceOracle are also attacked. Sry. Will clear that later.
---
tests the claimBuyerFunds function

-- tulip generation is not tests, since this is covered in the other tests


josojo:
I will make a new PR with cleaner structure.

PR #177, opened 2018-01-26
Contracts/smart contracts with overflow fix

W3stside:
@josojo 's PR

WIP

PR #174, opened 2018-01-26
Contracts/require solidity reinsert

W3stside:
Re-implementing require statements and running tests

1. removed `updateExchangeParams` in favor of cleaner funded approach

PR #178, opened 2018-01-27
Contracts/test/scwof with newest tests

josojo:
I added here all needed changes for all my tests priceOracle/historicalPriceOracle/claimSeller/claimBuyer

I also needed to update the smartcontract. Inject some getter functions for js and updated the getPrice function since it had again the stealing bug in it.


W3stside:
@josojo is pr #174 in here? if so will close

Velenir:
@W3stside not here, in #177

PR #179, opened 2018-01-29
testing for postBuyOrder.js

josojo:


test/dutchExchange-postBuyOrder.js
```diff
@@ -0,0 +1,215 @@
...
+      eth.address,
+      gno.address,
+      10.0.toWei(),
+      0,
+      2,
+      1,
+      { from: seller1 },
+    )
+    const auctionIndex = getAuctionIndex(eth, gno)
+    assertRejects(postBuyOrder(eth, gno, auctionIndex, 100, buyer1))
```
Velenir: Should be `await assertRejects()`. Here and in the last three tests

Velenir (approved):
Code looks very familiar, almost as if I've seen it in some other `post*Order` test, I wonder where...

For that alone, approve, right after `asserRejects` changes.

test/dutchExchange-postBuyOrder.js
Velenir: @josojo you missed this^ one, L150

PR #180, opened 2018-01-30
Added proxy functionality

josojo:
dom's new code with proxy

### DIFF
--- a/contracts/DutchExchange.sol
+++ b/contracts/DutchExchange.sol
@@ -102,6 +102,22 @@ contract DutchExchange {
     )
         public
     {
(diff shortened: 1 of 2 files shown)


contracts/DutchExchange.sol
```diff
@@ -19,6 +19,14 @@ contract DutchExchange {
         uint den;
     }
 
+    struct masterCopyCountdownType {
+        DutchExchange masterCopy;
+        uint timeWhenAvailable;
+    }
+
+    DutchExchange masterCopy;
```
josojo, line 27: Are you sure that we can define it as DutchExchange? Because in proxy contract, it is only an address.
I don't know it

josojo:
closing, since we get this code in PR 183

PR #182, opened 2018-01-30
Contracts/dutch exchange with proxy

josojo:
Here is the Proxy functionality with deployment and adaptions needed to make all tests talk to the right Proxy smart contracts.


josojo:
all this code is also in PR 183

PR #184, opened 2018-01-31
Contracts/test/tulip erc20 erc20

W3stside:
1. added non-Eth trade flow testing
2. changed checkTulipTokens testFunction to be simpler - now called assertUserReceivesTulips
3. changed the auctionIndex default from nothing to explicitly call tokens to check auc idx for


Velenir (approved):
Looks good, works even better!

PR #185, opened 2018-01-31
Contracts/test/proxy dx change

Velenir:
Tests for changing `DutchExchange` address in `Proxy` contract
The test logic is done and working, but checking `masterCopy` address needs to be done with `InternalTests` contract.


test/Proxy.js
```diff
@@ -0,0 +1,138 @@
...
+    log('tx was rejected')
+  })
+
+  it('auctioneer can update masterCopy after time limit', async () => {
+    assertIsAuctioneer(master)
+    const params1 = await getExchangeParams()
+
+    log('calling dx.updateMasterCopy() as auctioneer after time limit')
+    await dx.updateMasterCopy({ from: master })
+
```
josojo: I would add here one small assert.equal(dxNew.address,await dx.masterCopy.call())

but probably you have not done this because masteCopy is not public. Should we make it public in general?

Anyways this is just details. Good testing.

josojo (approved):

PR #189, opened 2018-02-01
added claimAndWithdraw test to claimSellerFunds

W3stside:
simple test - checks users token balance after claimAndWithdraw


josojo (approved):
thumbs up

Velenir (approved):

PR #190, opened 2018-02-02
added GNO-ETH claiming

W3stside:
1. Should take care of missing lines in `codecov`


josojo (approved):
lots of code to just touch 2 lines in solidity. 

Thanks for doing it quickly.

PR #191, opened 2018-02-03
some cleanings and bigger time variance allowance

josojo:
We still have some uncleanness in the code:
1. testing contracts are not in the testing files
2. Getter contract is needed 
3. some more documentation is needed


josojo:
replaced by PR from codeAudit

PR #192, opened 2018-02-05
Contracts/code audit

josojo:
We should also use the code from the audit as our base. Basically only cleaning and overflow fix is the difference.


josojo:
we decided against the removal of internaltestscripts from the deployment code

PR #193, opened 2018-02-06
Contracts/core audit1.0

josojo:
code clean for audit.


contracts/.solhint.json
```diff
@@ -1,4 +1,4 @@
-{                                              
+diud{
```
Velenir, line 1: what is this, where did it come from? mistyped?

josojo:
We will work with smartContractAudit

PR #194, opened 2018-02-06
Refactor/api/tokens

Velenir:
Changes to `api/Tokens.ts` and related types to bring them up to date with the contracts

Major changes -- two new functions:
* depositETH
* withdrawETH

only available for `TokenETH`, that's why they aren't generic like all others from `ERC20Interface`

Do we need any other `Token*` methods exposed like that in FE?


src/api/contracts.ts
```diff
@@ -9,14 +16,18 @@ const contractNames = [
   'Token',
   'TokenETH',
```
W3stside: should be `Ether Token` here

W3stside: applicable everywhere else as well below

src/api/index.ts
```diff
@@ -21,10 +21,11 @@ export const getAllAccounts = async () => {
 }
 
 // ether balance, not ETH tokens
-export const getETHBalance = async () => {
+export const getETHBalance = async (account?: Account) => {
```
W3stside: `getETHBalance` doesn't exist for me on the `web3` object. Is it not `web3.eth.getBalance(...)`?

src/api/types.ts
```diff
@@ -8,7 +8,7 @@ export type Index = number | BigNumber
 export interface ProviderInterface {
   getCurrentAccount(): Promise<Account>,
   getAccounts(): Promise<Account[]>,
-  getETHBalance(account?: Account): Promise<BigNumber>,
+  getETHBalance(account: Account): Promise<BigNumber>,
```
W3stside: ignore if previous comment is wrong but if not these need to change as well

src/api/types.ts
```diff
@@ -62,21 +65,82 @@ export type Filter = 'latest' | 'pending' | FilterObject | void
...
 export interface ERC20Interface {
   address: Account,
-  getTotalSupply(): Promise<BigNumber>,
-  balanceOf(account?: Account): Promise<BigNumber>,
-  transfer(to: Account, value: Balance, sender: Account, tx?: TransactionObject): Promise<Receipt>,
-  transferFrom(sender: Account, to: Account, value: Balance, tx?: TransactionObject): Promise<Receipt>,
-  approve(spender: Account, value: Balance, sender: Account, tx?: TransactionObject): Promise<Receipt>,
+  totalSupply(): Promise<BigNumber>,
+  balanceOf(account: Account): Promise<BigNumber>,
+  transfer(to: Account, value: Balance, tx?: TransactionObject): Promise<Receipt>,
```
W3stside: IMO - as these only return a `bool` based on transfer success or not, this, `transferFrom && approve` should have a manadatory `tx` param

W3stside (changes requested):

src/api/types.ts
Velenir: Yes, you're right

W3stside (approved):
Nice stuff, pls merge

PR #195, opened 2018-02-06
Feature/ts type 2.0 transition

W3stside:
1. Transition `DutchAuction` (`DutchExchange.sol` interface) into 2.0


src/api/types.ts
```diff
@@ -81,44 +81,77 @@ export interface Receipt {
...
+    auctioneer: Account,
+    ethUSDOracle: Account,
+    thresholdNewTokenPair: Balance,
+    thresholdNewAuction: Balance,
+    tx?: TransactionObject,
+  ): Promise<Receipt>,
   updateApprovalOfToken(token: Account, approved: boolean, tx?: TransactionObject): Promise<Receipt>,
-  updateETHUSDPriceOracle(_ETHUSDOracletoken: Account, tx?: TransactionObject): Promise<Receipt>,
+  startMasterCopyCountdown(masterCopy: Account, tx?: TransactionObject): Promise<Receipt>,
+  updateMasterCopy(tx?: TransactionObject): Promise<Receipt>,
```
Velenir: For some of these `tx` parameter isn't optional as `msg.sender` is checked inside the function.

W3stside: hmm thought I had changed that. Thanks

src/api/types.ts
```diff
@@ -155,22 +186,64 @@ export interface DXAuction {
...
+  computeRatioOfHistoricalPriceOraclesForJS(
+    tokenA: Account,
+    tokenB: Account,
+    auctionIndex: Index,
+  ): Promise<[BigNumber, BigNumber]>,
+
+  // internal
+  getTokenOrder(tokenA: Account, tokenB: Account): never,
+  setAuctionStart(tokenA: Account, tokenB: Account, value: number): never,
+  resetAuctionStart(tokenA: Account, tokenB: Account): never,
```
Velenir: `internal` methods are not exposed on the contract in JS, no need to add them

W3stside:
Failing Travis checks that are being fixed in incoming PR

Velenir (approved):

PR #197, opened 2018-02-07
refactor/api/high-level_Proposal

Velenir:
Some functions we will likely need in `api/index.ts`

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -21,10 +21,11 @@ export const getAllAccounts = async () => {
 }
 
 // ether balance, not ETH tokens
-export const getETHBalance = async () => {
+export const getETHBalance = async (account?: Account) => {
   const { web3 } = await promisedAPI
+  if (!account) account = await web3.getCurrentAccount()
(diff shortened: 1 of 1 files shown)


src/api/index.ts
```diff
@@ -153,14 +154,88 @@ export const getClaimedAmount = async (pair: TokenPair, index?: Index, account?:
...
+}
+
+/*
+ * manually checks if enough funds are allowed for depositing to DX,
+ * if not enough then sets allowance to amount required
+ * @param code TokenCode
+ * @param amount amount to have in allowance
+ * @param account 
+ */
+export const approveIfNotEnoughAllowance = async (code: TokenCode, amount: Balance, account?: Account) => {
```
W3stside, line 222: @Velenir not sure I like this one - seems dangerous and counterintuitive to the `allowance` idea

Velenir: ok, let's scrap it

Velenir:
Closed as outdated

PR #200, opened 2018-02-09
Feature/weth 2 eth

W3stside:
WIP *edit

Balance showed as `web3.getBalance(account)`  aka normal ETH

1. inserted W-ETH in `TokenCode` as typing
2. changed `getCurrentBalance` function to check for correct ETH

**edit -> will need auto conversion to W-ETH when selling / buying

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -31,14 +31,21 @@ export const getETHBalance = async (account?: Account) => {
 }
 
 // ETH token balance
-export const getCurrentBalance = async (account?: Account) => {
-  const { Tokens } = await promisedAPI
+export const getCurrentBalance = async (tokenName?: TokenCode | 'ETH', account?: Account) => {
+  account = await fillDefaultAccount(account)
+  
+  if (tokenName && tokenName === 'ETH') {
+    const { getETHBalance } = await promisedWeb3
 
+    return getETHBalance(account)
+  }
+
+  const { Tokens } = await promisedAPI
   // account would normally be taken from redux state and passed inside an action
   // but just in case
-  account = await fillDefaultAccount(account)
 
-  return Tokens.getTokenBalance('ETH', account)
+  // should probably change name here to WETH
+  return Tokens.getTokenBalance('WETH', account)
 }
 
 export const getTokenBalance = async (code: TokenCode, account?: Account) => {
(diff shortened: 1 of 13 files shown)


src/reducers/tokenBalances.ts
```diff
@@ -7,16 +7,16 @@ import { setTokenBalance } from 'actions/tokenBalances'
 // const zeroBalance = codeList.reduce((acc, code) => (acc[code] = 0, acc), {}) as TokenBalances
 
 // TODO: fill in when we have actions + fill in rest of Tokens + proper typing
-export default handleActions<TokenBalances | any>(
+export default handleActions<TokenBalances>(
   {
     [setTokenBalance.toString()]: (state, action) => ({
       ...state,
-      [action.payload.tokenName]: action.payload.balance,
+      [action.payload['tokenName']]: action.payload['balance'],
```
Velenir, line 14: why? TS fault?

W3stside: yeah weird error saying `tokenName` non existant on `TokenBalances` type

src/api/index.ts
```diff
@@ -31,14 +31,21 @@ export const getETHBalance = async (account?: Account) => {
...
+  }
 
+  const { Tokens } = await promisedAPI
   // account would normally be taken from redux state and passed inside an action
   // but just in case
-  account = await fillDefaultAccount(account)
 
-  return Tokens.getTokenBalance('ETH', account)
+  // should probably change name here to WETH
+  return Tokens.getTokenBalance('W-ETH', account)
```
Velenir, line 48: what about using `tokenName` here, maybe defaulting it to `W-ETH`

W3stside: that was my initial idea, and it worked but I'm honestly having second thoughts about this W-ETH thing. Rn I'm on the same PR and reverting back a commit. I think I want to keep the ETH name the same and just run a conditional there to check balance different if `tokenName` === `ETH`

src/reducers/tokenBalances.ts
Velenir: change to 
```
handleActions<TokenBalances, {tokenName: string, balance: string}>(
```

second temlate arg is payload

W3stside:
Prefer #201 over this

W3stside:
closing in favour of #201

PR #202, opened 2018-02-09
changedFeeFunction --> coreAudit1.0

hacker-DOM:


josojo:
I made all tests pass. We have no longer 100% coverage, but this is not important.

@dteiml please review the solidity code. I changed it, since your solution had a small issue. if (t==0), the correct solution was set but then again overwritten.

josojo:
merged into smartcontractaudit

PR #204, opened 2018-02-16
fns for JS --> ext fns

hacker-DOM:


hacker-DOM:
@josojo pls review & merge. thanks

PR #207, opened 2018-02-21
Feature/redirect from wallet page

Velenir:
Implements `RedirectHome` HOC redirecting to `/` page if user has not previously set `tokenPair.sellAmount`, which happens when user goes directly to `/wallet` page


W3stside (approved):
Looks good, works fine. Merge

PR #216, opened 2018-03-08
DX-132: first draft of added functions for frontend

josojo:
added the function as requested.

But there is no way for solidity to have int[][] in a dynamic or predefined way in memory. Hence functions are not as powerful/informative as they could be.

### DIFF
--- a/contracts/DutchExchange.sol
+++ b/contracts/DutchExchange.sol
@@ -1101,22 +1101,36 @@ contract DutchExchange {
         }
     }
 
-    // function getRunningTokenPairs(
-    //     address[] tokens
-    // )
-    //     public
-    //     view
-    //     returns (address[] tokens1, address[] token2)
-    // {
-    //     for (uint i = 0; i < tokens.length - 1; i++) {
-    //         for (uint j = i + 1; j < tokens.length; j++) {
-    //             if (getAuctionIndex(tokens[i], tokens[j]) > 0) {
-    //                 tokens1.push(tokens[i]);
-    //                 tokens2.push(tokens[j]);
-    //             }
-    //         }
-    //     }
-    // }
+    function getRunningTokenPairs(
+        address[] tokens
+    )
+        public
+        view
+        returns (address[] tokens1, address[] tokens2)
+    {
+        uint arrayLength;
+
+        for (uint k = 0; k < tokens.length - 1; k++) {
+            for (uint l = k + 1; l < tokens.length; l++) {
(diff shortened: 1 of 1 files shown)


contracts/DutchExchange.sol
```diff
@@ -1100,7 +1100,67 @@ contract DutchExchange {
...
+         address auctionBuyToken,
+        address user
+    )
+        public
+        view
+        returns(uint[])
+    {
+        uint runningAuctionIndex = getAuctionIndex(auctionSellToken, auctionBuyToken);
+        uint[] memory sellBal = new uint[](runningAuctionIndex+1);
+        for(uint j = 0; j <= runningAuctionIndex; j++){
```
W3stside: A zero index pair can't be possible, so this should be `j = 1` no?

josojo: in js we would have to put j=1, but in solidity j=0 should be fine

W3stside: 👍

contracts/DutchExchange.sol
```diff
@@ -1100,7 +1100,67 @@ contract DutchExchange {
...
+    //@param auctionSellToken are the sellTokens defining an auctionPair
+    //@param auctionBuyToken are the buyTokens defining an auctionPair
+    //@param user is the user who wants to his tokens
+    function getSellerBalanceOfCurrentAuctions(
+        address[] auctionSellToken,
+        address[] auctionBuyToken,
+        address user
+    )
+        public
+        returns(uint[])
```
Velenir: Should this also be a `view` function?And the next one

hacker-DOM: they should

contracts/DutchExchange.sol
hacker-DOM: I agree with @W3stside that it should begin at j = 1

contracts/DutchExchange.sol
```diff
@@ -1101,6 +1101,121 @@ contract DutchExchange {
...
+        address[] auctionSellTokens,
+        address[] auctionBuyTokens,
+        address user
+    )
+        public
+        view
+        returns (uint[])
+    {
+        uint length = auctionSellTokens.length;
+        uint length2 = auctionBuyTokens.length;
```
W3stside, line 1193: would setting `length2` here back to 0 save gas? or am I being dumb 😓

josojo: hopefully we will never have to pay any gas for the constant view functions :)

PR #222, opened 2018-04-05
Feature/DX-173-IPFS

W3stside:
- [x] Use uploaded Tokens.JSON in IPFS / defaultJSON as data to populate Tokens to choose
- [x] Add way to grab `defaultJSON`
- [x] Figure out what address to default to


W3stside:
Forgot to mention, WIP

src/api/utils/index.ts
```diff
@@ -13,6 +13,12 @@ export const windowLoaded = new Promise((accept, reject) => {
   }, false)
 })
 
+export const readFileUpload = (file: any) =>
```
Velenir: Using native types this can be `export const readFileUpload = (file: File): Promise<ArrayBuffer> =>`

src/types/index.d.ts
```diff
@@ -110,14 +110,33 @@ export interface TokenOverlay {
   mod: TokenMod
 }
 
+export type FileBuffer = number[]
```
Velenir: Why is it `number[]`. It's what you get after `FileReader.readAsArrayBuffer`? Then it's `ArrayBuffer` type

src/types/index.d.ts
```diff
@@ -110,14 +110,33 @@ export interface TokenOverlay {
...
+export interface oFile {
```
Velenir: This can be the native `File` type, no? Interface will remain the same

src/api/utils/index.ts
W3stside: done

src/types/index.d.ts
W3stside: done

src/types/index.d.ts
W3stside: done

src/containers/TokenUpload/index.ts
```diff
@@ -1,32 +1,57 @@
...
 
-import { setUploadFileParams, setIPFSFileHashAndPath, getFileContentFromIPFS, openModal } from 'actions'
+import { setUploadFileParams, setIPFSFileHashAndPath, getFileContentFromIPFS, openModal, setCustomTokenList } from 'actions'
 
 import IPFSHOC from 'components/IPFSHOC'
 import TokenUpload from 'components/TokenUpload'
 
 import { State } from 'types'
 
+import { batchActions } from 'redux-batched-actions'
```
W3stside: in here as I felt too lazy to move it elsewhere in case we didnt like it -___-

src/reducers/tokenList.ts
```diff
@@ -1,20 +1,39 @@
...
+        ...action.payload.defaultTokenList,
+      ],
     }),
     [setCustomTokenList.toString()]: (state: any, action: any) => ({
       ...state,
       customTokenList: action.payload.customTokenList,
+      combinedTokenList: [
+        ...state.defaultTokenList,
+        ...state.customTokenList,
+        ...action.payload.customTokenList,
```
Velenir: this will always add to the current `combinedTokenList`. Try uploading the same file twice and see the duplicates appear.
Removing `...state.customTokenList` will fix that.

Velenir (changes requested):

Velenir (approved):

PR #226, opened 2018-04-13
Feature/dx 231 generates mgn

Velenir:
Shows that **TokenPair** in **OrderPanel**  won't generate MGN if any token isn't approved
Shows that a token isn't approved inside **TokenList** in **TokenOverlay**

For now uses a mini test list of approved tokens `[ETH, GNO]` only.
Will need to change some logic in `getDefaultTokens` from `ipfs` branch, when that is finalized and merged.
Will need even more changes (fetching, storage in localForage?) when `whitelist/approvedTokens` list is available.


src/containers/OrderPanel/index.ts
```diff
@@ -2,11 +2,19 @@ import { connect } from 'react-redux'
 import OrderPanel from 'components/OrderPanel'
 import { State } from 'types'
 
-const mapStateToProps = ({ tokenPair: { sell, buy, sellAmount } }: State) => ({
-  sellToken: sell,
-  buyToken: buy,
-  sellAmount,
-})
+const isTokenApproved = ({ approvedTokens, tokenPair: { sell, buy } }: State) =>
```
W3stside: love how clean this is, nice

src/integrations/walletIntegration.ts
```diff
@@ -18,6 +18,11 @@ export default async function walletIntegration(store: Store<any>) {
     registerProvider: dispatchProviderAction(registerProvider),
   }
 
+  // TODO: fetch approvedTokens list from api
+  // then after getting tokensJSON in getDefaultTokens create a list of approved TokenCodes
+  // then only dispatch that list
+  dispatch(setApprovedTokens(['ETH', 'GNO']))
```
W3stside: this is only `dispatch`ing the `symbols` yet you check the whole `TokenObject` in `isTokenApproved`

Velenir: Why is it here? This is old, I don't have it locally

W3stside (approved):

PR #232, opened 2018-04-24
Feature/dx 233 blocking

Velenir:
Blocks App through rendering only static html
Conditions for blocking are extendable


W3stside (approved):
Very nice

PR #234, opened 2018-04-25
Feature/dx 252 modal tx details

W3stside:
1. added `txData` prop to `modalProp` to pass more info during Transactions
2. Moved JS CSS into `modals/_modals.scss` > `globals.scss`
3. Wording changes + types

_to test_: 
1. `trufflescripts/add_token_pair.js --pair eth,gno`
2. submit sell order
3. wrap eth, approve etc
4. Modal for `Sell Confirmation` is changed Modal


src/components/Modals/index.tsx
```diff
@@ -18,76 +18,61 @@ interface BlockModalProps extends TransactionModalProps {
...
+  <div className="modalDivStyle">
+    <h1>{header || 'Approving Tokens and Transferring'}</h1>
+    <p className="modalHeaderDescriptor">
       { body || `Please check your ${activeProvider || 'Provider'} notifications in extensions bar of your browser.` }
     </p>
-    {error && 
+    {txData && <div className="modalTXDataDiv">
+      <ul>
+        <li>{`Sell Amount:    ${txData.sellAmount}`}</li>
+        {(txData.tokenA.symbol || txData.tokenA.symbol) &&
```
Velenir: This should be `txData.tokenA.symbol || txData.tokenA.name` right?
What about address when there is no `symbol or name`?

Velenir (dismissed):

src/components/Modals/index.tsx
W3stside: `address` is below anyways - figured dont render the li at all if there is no `name` or `symbol`. but good catch on the other!

W3stside: done

Velenir (approved):

PR #235, opened 2018-04-26
Feature/dx 253 loading component

W3stside:
# WIP

Adds `renderProp` `Loading` component using `Aux` as passthrough when necessary

- renders conditionally based on pass in `data` prop

Needs more work. Needs to be placed in relevant places

_To Test:_ Head to `dxURL/auction/eth-gno-5` (arbitrarily high number here)

Also as a bonus: `npm run addTokenPairs` to add `gno, omg, rdn` and sim trading (doesn't start auctions)

Yay SVG animations!


Velenir (approved):
What a pretty animation

Also I was thinking would it make sense to combine `npm run migrate && npm run addTokenPair` into some `setup-rpc` script?

PR #236, opened 2018-04-26
Feature/dx 254 default token svg

W3stside:
# WIP

Need to wait for official answer on images + Michel's rendition of `defaultToken.svg` if approved


Velenir (approved):
default token ftw

PR #238, opened 2018-04-27
Hotfix/custom token list wait

W3stside:
# Fixes

1. customTokenList on initial load was not properly grabbing new `tokenBalances` for newly added tokens and breaking `React` render
  a. fixed this to `async` load `getTokenBalances` after setting new list and showing loader in `tokenOverlay` while tokens load in

2. Put SVG in container that is resizable via props to make it more useable elsewhere
3. Added `message` prop to SVG to add custom messages while loading


src/actions/tokenList.ts
```diff
@@ -1,17 +1,34 @@
...
+  ], 'BATCH_SET_CUSTOM_LIST_AND_IPFS_HASH_PATH'))
+
+  // grab updated customTokenList from State
+  const { tokenList: { customTokenList } } = getState()
+
+  // set list type to Custom
+  dispatch(setTokenListType({ type: 'CUSTOM' }))
+
+  // recalculate Token Balances
+  const tokenBalances = await getTokenBalances(customTokenList)
```
Velenir, line 28: good idea checking balances when token list changes in addition to in `updateState`

Velenir (approved):

PR #239, opened 2018-04-28
Feature/dx 255 nowallet when mm locked

Velenir:
- [x] `NoWallet` is dsiplayed when Mmask is locked or unavailable
- [x] switches between states on lock/unlock

- [x] display different messages based on locked/available state
- [ ] remove `Loading...` in the Header


W3stside:
Current changes approved

W3stside (approved):

PR #242, opened 2018-05-03
Feature/dx 233 blocking codes

Velenir:
Added blocking codes in a separate file
Fix for wallet availability detection flipping `walletEnabled` prop repeatedly


W3stside (approved):
Looks good to me

PR #245, opened 2018-05-04
Fix/dx 268 no sell amount more than sell balance

Velenir:
Doesn't allow to proceed to `WalletPanel` when `sellAmount> sellTokenBalance`
Also displays native message `amount available for sale is ${maxSellAmount}` at input


Velenir:
Closed due to strange local git conflict
Same as #246

PR #246, opened 2018-05-04
Fix/dx 268 sell amount

Velenir:
Doesn't allow to proceed to `WalletPanel` when `sellAmount> sellTokenBalance`
Also displays native message `amount available for sale is ${maxSellAmount}` at input


W3stside (approved):
super nice, approved

PR #240, opened 2018-04-30
changed formatting of reciprocal auctions display + removed Index Column

W3stside:
# Ongoing Auctions Rework - Claiming

1. Formatted display of reciprocal auctions differently
- [x] ~Show `CLEARED` status in `Committed` `<th>` instead of `sellVolume` on auction clear~
- [x] Insert logic for not removing auction if cleared and has unclaimed funds
- [x] Replace `CLEARED` with newly committed amounts if new auction for same pair has started and funds have not been claimed

PR #247, opened 2018-05-05
EXTENSION: Feature/dx 256 ongoing auction claiming and linking

W3stside:
# Extends #240 

1. Added `Link` component to move user to `AuctionStatus` page related to `ongoingAuction` clicked
2. Created `claimSellerFundsFromSeveral` thunk/ `api/index` and `api/dutchx` to enable claiming from auctions with sellerbalances and clear status
3. removed state print in `AuctionStatusHOC`

### DIFF
--- a/src/actions/blockchain.ts
+++ b/src/actions/blockchain.ts
@@ -359,8 +359,11 @@ export const claimSellerFundsFromSeveral = (
         body: `Claiming ${buyName} tokens from ${sellName}-${buyName} auction. Please check ${activeProvider}`,
       },
     }))
-    const claimReceipt = await getUnclaimedSellerFundsFromSeveral(sell, buy, currentAccount, indices)
-    console.log('​claimReceipt => ', claimReceipt)
+    const [claimReceipt] = await Promise.all([
+      getUnclaimedSellerFundsFromSeveral(sell, buy, currentAccount, indices),
+      updateMainAppState(),
+    ])
+    console.log('​Claim receipt => ', claimReceipt)
     return dispatch(closeModal())
   } catch (error) {
     console.error(error.message)
--- a/src/actions/blockchain.ts
+++ b/src/actions/blockchain.ts
@@ -19,6 +19,7 @@ import {
   getTokenBalance,
   toNative,
   claimSellerFundsFromSeveralAuctions,
+  getIndicesWithClaimableTokensForSellers,
 } from 'api'
 
 import {
@@ -69,13 +70,15 @@ export const setTokenSupply = createAction<{ mgnSupply: string | BigNumber }>('S
 
 const NETWORK_TIMEOUT = process.env.NODE_ENV === 'production' ? 10000 : 200000
 
-export const updateMainAppState = () => async (dispatch: Function, getState: () => State) => {
+export const updateMainAppState = (condition?: any) => async (dispatch: Function, getState: () => State) => {
   const { tokenList } = getState()
   const mainList = tokenList.type === 'DEFAULT' ? tokenList.defaultTokenList : tokenList.combinedTokenList
   const [{ TokenMGN }, currentAccount] = await Promise.all([
     promisedContractsMap,
     getCurrentAccount(),
   ])
+
+  const status = condition.fn && typeof condition.fn === 'function' ? condition.fn && await condition.fn(...condition.args) : condition
 
   // Check state in parallel
   /*
@@ -107,6 +110,8 @@ export const updateMainAppState = () => async (dispatch: Function, getState: () 
     setFeeRatio({ feeRatio: feeRatio.toNumber() }),
     setTokenSupply({ mgnSupply: mgn.balance.toString() }),
   ], 'HYDRATING_MAIN_STATE'))
+
+  return status
 }
 
 // CONSIDER: moving this OUT of blockchain into index or some INITIALIZATION action module.
@@ -359,11 +364,16 @@ export const claimSellerFundsFromSeveral = (
         body: `Claiming ${buyName} tokens from ${sellName}-${buyName} auction. Please check ${activeProvider}`,
       },
     }))
-    const [claimReceipt] = await Promise.all([
-      claimSellerFundsFromSeveralAuctions(sell, buy, currentAccount, lastNIndex),
-      updateMainAppState(),
-    ])
+    const claimReceipt = await claimSellerFundsFromSeveralAuctions(sell, buy, currentAccount, lastNIndex)
     console.log('​Claim receipt => ', claimReceipt)
+
+    // refresh state ...
+    let [, sellBalance] = await dispatch(updateMainAppState({ fn: getIndicesWithClaimableTokensForSellers, args: [{ sell, buy }, currentAccount, 0] }))
+    // loop until sellBalance drops to 0
+    while (sellBalance.length && sellBalance[0].gt(0)) {
+      ([, sellBalance] = await dispatch(updateMainAppState({ fn: getIndicesWithClaimableTokensForSellers, args: [{ sell, buy }, currentAccount, 0] })))
+    }
+
     return dispatch(closeModal())
   } catch (error) {
     console.error(error.message)
--- a/src/actions/blockchain.ts
+++ b/src/actions/blockchain.ts
@@ -384,7 +384,7 @@ async function checkTokenAllowance(
   return tokenAllowance
 }
 
-function errorHandling(error: Error) {
+export function errorHandling(error: Error) {
   const errorFind = (string: string, toFind = '}', offset = 1) => {
     const place = string.search(toFind)
     return string.slice(place + offset)
--- a/src/actions/blockchain.ts
+++ b/src/actions/blockchain.ts
@@ -18,6 +18,7 @@ import {
   getETHBalance,
   getTokenBalance,
   toNative,
+  getUnclaimedSellerFundsFromSeveral,
 } from 'api'
 
 import {
@@ -37,6 +38,7 @@ import { timeoutCondition } from '../utils/helpers'
 import { BigNumber, TokenBalances, Account, State } from 'types'
 import { promisedContractsMap } from 'api/contracts'
 import { DefaultTokenObject } from 'api/types'
+import { Dispatch } from 'react-redux'
 
 export enum TypeKeys {
   SET_GNOSIS_CONNECTION = 'SET_GNOSIS_CONNECTION',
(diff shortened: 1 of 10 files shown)


src/components/MenuAuctions/index.tsx
```diff
@@ -26,13 +30,17 @@ export const MenuAuctions: React.SFC<MenuAuctionProps> = ({
               {ongoingAuctions.map(
                 (auction, i) =>
                   <tr key={`${auction.sell.address}-${auction.buy.address}-${i}`}>
-                    <td>{`${auction.sell.symbol}/${auction.buy.symbol}`}</td>
+                    <td>
+                      <Link to={`/auction/${auction.sell.symbol}-${auction.buy.symbol}-${auction.indices[auction.indices.length - 1]}`}>
```
Velenir: This `Link` is always to the latest (ongoing) auction, what if user doesn't have anything commited in that auction, only claimable in past auctions? Where should the link lead?

src/api/index.ts
```diff
@@ -533,31 +562,35 @@ export const getSellerOngoingAuctions = async (
...
 
       const { sell: { decimals }, buy: { decimals: decimalsInverse } } = auction
-      const { lastIndex, closingPriceDir, closingPriceOpp } = lastAuctionsData[index]
-
-      const ongoingAuction = {
-        ...auction,
-        indices,
-        indicesInverse,
+      if (!(indices.length >= 1 || indicesInverse.length >= 1)) {
+        return accum
```
Velenir: why the change? it does literally the same, but creates `ongoingAuction` with let, unnecessarily, has  potentially skippable on the first conditional `auction` destructuring, unnecessarily, uses `else` instead of an early `return`

src/api/index.ts
```diff
@@ -363,6 +363,34 @@ export const getUnclaimedSellerFunds = async (pair: TokenPair, index?: Index, ac
...
+  buy: DefaultTokenObject,
+  userAccount?: Account,
+  indices: number = 0,
+) => {
+  const { DutchX } = await promisedAPI
+  userAccount = await fillDefaultAccount(userAccount)
+  const sellArr = [], buyArr = []
+
+  const claimableIndices = (await DutchX.getIndicesWithClaimableTokensForSellers(sell.address, buy.address, userAccount, indices))[0].map((i: BigNumber) => i.toNumber())
+  const claimableIndicesBalance = (await DutchX.getIndicesWithClaimableTokensForSellers(sell.address, buy.address, userAccount, indices))[1].map((i: BigNumber) => i.toNumber())
```
Velenir: This and above can be inside `Promise.all`. Also is `claimableIndicesBalance` only used for logging?

src/api/dutchx.ts
```diff
@@ -92,19 +92,32 @@ async function init(): Promise<DutchExchange> {
...
+    account: Account,
+  ) => dx.claimTokensFromSeveralAuctionsAsSeller(
+    sellTokenAddresses,
+    buyTokenAddresses,
+    indices,
+    account,
+    { from: account },
+  )
+
+  const claimBuyerFunds = (
```
Velenir, line 110: Do we need to use `claimBuyerFunds` anywhere or is it here for completion's sake?

src/actions/blockchain.ts
```diff
@@ -341,6 +343,32 @@ export const approveAndPostSellOrder = (choice: string) => async (dispatch: Func
   }
 }
 
+export const claimSellerFundsFromSeveral = (
+  sell: DefaultTokenObject,
+  buy: DefaultTokenObject,
+  indices?: number,
```
Velenir: Can you rename `indices` to something like `lastNIndex`. I was expecting it to be an array based on the name...

src/api/index.ts
```diff
@@ -363,6 +363,34 @@ export const getUnclaimedSellerFunds = async (pair: TokenPair, index?: Index, ac
   }
 }
 
+export const getUnclaimedSellerFundsFromSeveral = async (
```
Velenir: This function actually claims the tokens, not just shows how much is claimable? The `get` in the name confuses me.

src/api/index.ts
```diff
@@ -363,6 +363,34 @@ export const getUnclaimedSellerFunds = async (pair: TokenPair, index?: Index, ac
...
+  sell: DefaultTokenObject,
+  buy: DefaultTokenObject,
+  userAccount?: Account,
+  indices: number = 0,
+) => {
+  const { DutchX } = await promisedAPI
+  userAccount = await fillDefaultAccount(userAccount)
+  const sellArr = [], buyArr = []
+
+  const claimableIndices = (await DutchX.getIndicesWithClaimableTokensForSellers(sell.address, buy.address, userAccount, indices))[0].map((i: BigNumber) => i.toNumber())
```
Velenir: I know it shouldn't happen based on the place the function is called, but please add a check like:
```
if (claimableIndices .length === 0) return
```
or some such.

src/api/index.ts
```diff
@@ -363,6 +363,34 @@ export const getUnclaimedSellerFunds = async (pair: TokenPair, index?: Index, ac
...
+  const claimableIndicesBalance = (await DutchX.getIndicesWithClaimableTokensForSellers(sell.address, buy.address, userAccount, indices))[1].map((i: BigNumber) => i.toNumber())
+
+  const lastIndexCleared = ((await DutchX.getClosingPrice({ sell, buy }, claimableIndices[claimableIndices.length - 1])).map(i => i.toNumber()))[0] > 0
```
Velenir: Can be one action shorter with
```
(await DutchX.getClosingPrice({ sell, buy }, claimableIndices[claimableIndices.length - 1])[0].greater(0)
```
no need to `.map`

src/api/index.ts
```diff
@@ -363,6 +363,34 @@ export const getUnclaimedSellerFunds = async (pair: TokenPair, index?: Index, ac
...
+
+  console.log('claimableIndices => ', claimableIndices, 'claimableIndicesBalance => ', claimableIndicesBalance, 'lastIndexCleared? => ', lastIndexCleared)
+
+  for (let i = 1; i <= claimableIndices.length; ++i) {
+    if (!lastIndexCleared) continue
```
Velenir: You want to include `sell/buy.address` for all indices except for the last if it happens to be the ongoing auction not yet closed, right? `if (!lastIndexCleared) continue` will end up in empty arrays should it trigger.

What about:
```
cost length = lastIndexCleared ? claimableIndices.length : claimableIndices.length - 1
const sellArr = Array.from({ length }, () => sell.address)
const buyArr = ...
```
This will give for
```
claimableIndices [1,3,5] and lastIndexCleared = true  =>  sellArr = [address, address, address]

claimableIndices [1,3,5] and lastIndexCleared = false =>  sellArr = [address, address]
```

Velenir (changes requested):
Good stuff, much awaited claiming functionality

src/components/MenuAuctions/index.tsx
W3stside: maybe latest with claiming if current has no committed?

src/api/index.ts
W3stside: artifact from tests I am running for showing the first auction in `ongoingAuctions` if auction hasn't started yet.

consider the current UX now - on rinkeby or anywhere else - where a user commits funds to auction 1 that hasn't started. `ongoingAuctions` shows nothing, which is unclear and quite frankly, bad UX. We need to show something for the first auction index. So I changed the code here for just easier testing in order to at the very least return the auction pair returned from `getRunningPairs` and show `committed` amount

src/api/index.ts
W3stside: ok will change

src/api/index.ts
W3stside: will do

src/api/index.ts
W3stside: true

src/api/index.ts
W3stside: cool, nice optimization. will change

src/api/index.ts
W3stside: yes, artifact

src/api/dutchx.ts
W3stside: as of now, completion

src/actions/blockchain.ts
W3stside: yes, also an artifact (lots of those from weekend coding, lol) as before I did not check using `getIndicesWithClaimableTokensForSellers`

src/api/index.ts
W3stside: it claims. it's also the same name as in `DX.sol`

W3stside:
weeeekkkeennnddd coooodddiiinnngggg, this is the most you've wrecked my PR 😂 😂 😂 

all good stuff, thanks for the thorough checkthrough

src/components/MenuAuctions/index.tsx
W3stside: @Velenir are we ok with what I suggested?

src/actions/blockchain.ts
W3stside: done

src/components/MenuAuctions/index.tsx
Velenir: Yes, sounds reasonable

src/actions/blockchain.ts
```diff
@@ -359,8 +359,11 @@ export const claimSellerFundsFromSeveral = (
         body: `Claiming ${buyName} tokens from ${sellName}-${buyName} auction. Please check ${activeProvider}`,
       },
     }))
-    const claimReceipt = await getUnclaimedSellerFundsFromSeveral(sell, buy, currentAccount, indices)
-    console.log('​claimReceipt => ', claimReceipt)
+    const [claimReceipt] = await Promise.all([
+      getUnclaimedSellerFundsFromSeveral(sell, buy, currentAccount, indices),
+      updateMainAppState(),
```
Velenir: `updateMainState` should be fired after `getUnclaimedSellerFundsFromSeveral` tx is executed and mined, otherwise it can fetch the old state. And even then it won't necessarily bring in the new state as happened in owl project (MMask is to blame?)
You don't need to `await` on `updateMainState`, you can just start it after receipt is received and it'll run it's course by itself without stalling `claimSellerFundsFromSeveral`  function execution

Velenir (changes requested):

src/actions/blockchain.ts
W3stside: good call - also yes, I checked this in the app and it doesn't update appropriately

Velenir (approved):

PR #248, opened 2018-05-07
Feature/dx 248 block netw

Velenir:
Blocks UI when not run on `localhost` and on networks other than Rinkeby.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -31,6 +31,31 @@ const isGeoBlocked = async () => {
   }
 }
 
+const isNetBlocked = async () => {
+  if (typeof window === 'undefined') return false
+  const { hostname } = window.location
+  // allow anything when run locally
+  if (hostname === 'localhost' || hostname === '0.0.0.0') return false
+
+  // no walletextension detected, different error - download wallet error
+  if (!window.web3) return false
+
+  try {
+    const id = await new Promise((res, rej) => {
+      window.web3.version.getNetwork((e: Error, r: string) => e ? rej(e) : res(r))
+    })
+    // allow Rinkeby and local testrpc (id = Date.now())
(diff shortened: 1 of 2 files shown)


src/index.tsx
```diff
@@ -31,6 +31,31 @@ const isGeoBlocked = async () => {
   }
 }
 
+const isNetBlocked = async () => {
+  if (typeof window === 'undefined') return false
+  const { hostname } = window.location
+  // allow anything when run locally
+  if (hostname === 'localhost' || hostname === '0.0.0.0') return false
```
W3stside, line 38: i would just make this a production check - shouldnt be in production code

src/index.tsx
```diff
@@ -31,6 +31,31 @@ const isGeoBlocked = async () => {
...
+  // no walletextension detected, different error - download wallet error
+  if (!window.web3) return false
+
+  try {
+    const id = await new Promise((res, rej) => {
+      window.web3.version.getNetwork((e: Error, r: string) => e ? rej(e) : res(r))
+    })
+    // allow Rinkeby and local testrpc (id = Date.now())
+    //                      Apr 29 2018
+    if (id === '4' || id > 1525000000000) return false
```
W3stside, line 49: throw this in production check too

W3stside (approved):

src/index.tsx
Velenir: But what about https://dx.dev.gnosisdev.com/ which is `development` but entirely working?

src/index.tsx
Velenir: same as above

src/index.tsx
W3stside: ok true

PR #249, opened 2018-05-07
Fix/scss img urls in biuld

Velenir:
Fixes for relative image urls that change during build process


W3stside (approved):
if `dist` works, merge

PR #251, opened 2018-05-08
Hotfix/token decimals menu wallet

W3stside:
# hotfix + feature hybrid ... aka ... HOT FEATURE

1. fixes `getDecimals` where a zero `BigNumber` return value was causing `NaN` division in `MenuWallet` - now converts `toNumber` and `||`s to `18` decimals if `0`
2. Don't show tokens that have a balance `!>0`

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -98,7 +98,8 @@ export const getTokenDecimals = async (tokenAddress: Account) => {
   const { Tokens } = await promisedAPI
 
   try {
-    await Tokens.getTokenDecimals(tokenAddress)
+    const decimals = (await Tokens.getTokenDecimals(tokenAddress)).toNumber() || 18
+    return decimals
   } catch (e) {
     console.warn(`Token @ address ${tokenAddress} has no Decimals value set - defaulting to 18`)
     return 18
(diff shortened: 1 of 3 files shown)


src/api/index.ts
```diff
@@ -98,7 +98,8 @@ export const getTokenDecimals = async (tokenAddress: Account) => {
   const { Tokens } = await promisedAPI
 
   try {
-    await Tokens.getTokenDecimals(tokenAddress)
+    const decimals = (await Tokens.getTokenDecimals(tokenAddress)).toNumber() || 18
```
Velenir, line 101: The `|| 18` is useless, `getTokenDecimals()` either breaks and goes to catch or returns something and works.
It'll go to `|| 18` only if `getTokenDecimals().toNumber() == 0`. Is that ever possible even?

`return (await Tokens.getTokenDecimals(tokenAddress)).toNumber()` should be enough

Velenir (approved):

src/api/index.ts
Velenir: You know what, it is possible. Then all `balance.div(10 ** 0) === balance`. 0 is valid decimals, meaning no decimals, shouldn't override it with 18

W3stside: its not

W3stside: I was getting BN(0) from await get TokenDecimals w/custom EOS token

PR #252, opened 2018-05-08
Update/dutch exchange contracts

W3stside:
# Updated DX-Contracts and contract paths

... yep, that's pretty much it


Velenir (dismissed):
need to change `npm run rpc` and other scripts

Velenir (approved):

PR #253, opened 2018-05-08
Feature/new auction statuses

Velenir:
New auction statuses and conditions for checking auction state

### DIFF
--- a/src/components/AuctionPanel/index.tsx
+++ b/src/components/AuctionPanel/index.tsx
@@ -29,6 +29,7 @@ const AuctionPanel: React.SFC<AuctionPanelProps> = ({
   sell, buy,
   status, completed, timeToCompletion,
   userSelling, userGetting, userCanClaim,
+  error,
 }) => (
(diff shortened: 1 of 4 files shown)


W3stside:
@Velenir heading to an token pair auction index that is `PLANNED` stage has confusing aspects.

Shows `deposit confirmed` in status bar when that is sometimes not true - I would remove this via a conditional check or something else -  maybe we can add more concise wording to describe what is happening

_To Test_: `npm run addTokenPairs` -> head to `.../auction/ETH-GNO-2`

Velenir:
Yes, progress bar still basically counts steps between statuses [STATUS.PLANNED--ACTIVE--ENDED]. As it was at the very beginning.
That needs to be changed, but I'd like to dedicate a separate PR to that later.

W3stside:
@Velenir if you make the ticket and assign yaself I'll approve it :D

W3stside (approved):

PR #254, opened 2018-05-08
Fix/max sell amount

Velenir:
- Small float number for `sellAmount` doesn't trigger `amount available for sale is ...` alert
- `step="0.0001"` for `sellAmount` input


W3stside (approved):
solid, merge

PR #255, opened 2018-05-08
Feature/dx 273 ipfs download

W3stside:
# From Token uploading > Token list IPFS downloading

1. `TokenUpload` changed to work with new logic. Needs a name change probably.
2. Removed unncessarry and bulky state + fns that are no longer being used
3. changed styling to fit new `input` `type=text`

For a test file to grab: `QmZjVUBmJCnEytBycP8dciDbKCQje7tcZXvHn1RjsqcuDd` (may be gone)

To add file via `CLI`:
1. Tab1: `jsipfs daemon`
_Navigate to folder containing `tokens.json`_
2. Tab2: `jsipfs add tokens.json` (will return fileHash if successful)
_To check if u can cat file_
(optional) `jsipfs cat fileHashFromStep2`

What needs to change:
- [x] Logic if incorrect HASH is input in text and download is clicked - as of now nothing happens - `async` logic does NOT throw via `IPFSHOC` in `handleGrabFromIPFS` in function `const fileContent = await ipfsGetAndDecode(fileHash)` <- this should do something but it doesnt.
- [x] Bug checking around logic
- [x] Optimization

Have a good month bro! If you need anything, just let me know, I am on holiday but I'm not a dick and don't want you to feel stressed/shitty/helpless my man. Just message me if u need anything and I'll try to get back to you asap.

### DIFF
--- a/src/actions/tokenList.ts
+++ b/src/actions/tokenList.ts
@@ -1,7 +1,7 @@
 import { createAction } from 'redux-actions'
 import { batchActions } from 'redux-batched-actions'
 
-import { setTokenBalance, setIPFSFileHashAndPath } from 'actions'
+import { setTokenBalance } from 'actions'
 import { getTokenBalances } from 'api'
 
 import { TokenListType, State } from 'types'
@@ -12,11 +12,8 @@ export const setCustomTokenList = createAction<{ customTokenList: DefaultTokenOb
 
 export const setTokenListType = createAction<{ type: TokenListType['CUSTOM' | 'DEFAULT' | 'UPLOAD'] }>('SET_TOKEN_LIST_TYPE')
 
-export const setNewIPFSCustomListAndUpdateBalances = (p1: { customTokenList: DefaultTokenObject[] }, p2: {}) => async (dispatch: Function, getState: () => State) => {
-  dispatch(batchActions([
-    setCustomTokenList(p1),
-    setIPFSFileHashAndPath(p2),
-  ], 'BATCH_SET_CUSTOM_LIST_AND_IPFS_HASH_PATH'))
+export const setNewIPFSCustomListAndUpdateBalances = (p1: { customTokenList: DefaultTokenObject[] }) => async (dispatch: Function, getState: () => State) => {
+  dispatch(setCustomTokenList(p1))
 
   // grab updated customTokenList from State
   const { tokenList: { customTokenList } } = getState()
--- a/src/components/IPFSHOC/index.tsx
+++ b/src/components/IPFSHOC/index.tsx
@@ -35,7 +35,7 @@ export default (WrappedComponent: React.SFC<any> | React.ComponentClass<any>) =>
     handleFileHashInput = ({ target: { value } }: any) => this.props.setIPFSFileHash(value)
 
     handleGrabFromIPFS = async () => {
-      const { fileHash, openModal, setNewIPFSCustomListAndUpdateBalances } = this.props
+      const { fileHash, openModal, setIPFSFileHash, setNewIPFSCustomListAndUpdateBalances } = this.props
 
       try {
         const { ipfsGetAndDecode } = await promisedIPFS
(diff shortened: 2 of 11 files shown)


src/actions/tokenList.ts
```diff
@@ -12,11 +12,8 @@ export const setCustomTokenList = createAction<{ customTokenList: DefaultTokenOb
 
 export const setTokenListType = createAction<{ type: TokenListType['CUSTOM' | 'DEFAULT' | 'UPLOAD'] }>('SET_TOKEN_LIST_TYPE')
 
-export const setNewIPFSCustomListAndUpdateBalances = (p1: { customTokenList: DefaultTokenObject[] }, p2: {}) => async (dispatch: Function, getState: () => State) => {
-  dispatch(batchActions([
-    setCustomTokenList(p1),
-    setIPFSFileHashAndPath(p2),
-  ], 'BATCH_SET_CUSTOM_LIST_AND_IPFS_HASH_PATH'))
+export const setNewIPFSCustomListAndUpdateBalances = (p1: { customTokenList: DefaultTokenObject[] }) => async (dispatch: Function, getState: () => State) => {
```
Velenir, line 15: does it need to be a thunk? it doesn't use state or have any async logic

src/containers/TokenUpload/index.ts
```diff
@@ -9,32 +14,26 @@ import { State } from 'types'
...
   },
 }: Partial<State>) => ({
-  fileBuffer,
-  oFile,
   fileHash,
   filePath,
-  fileContent,
   json,
   customTokenList,
+  potentiallyValidHash: fileHash.length == 46,
```
Velenir: `===`?

src/components/IPFSHOC/index.tsx
```diff
@@ -1,128 +1,77 @@
...
+        this.setState({ pullingData: false })
 
         // setState
-        return getFileContentFromIPFS({ fileContent })
+        return setNewIPFSCustomListAndUpdateBalances({ customTokenList: customTokenListWithDecimals })
       } catch (error) {
         console.error(error)
+        // reset hash to empty string
+        setIPFSFileHash('')
+
```
Velenir: you never change state on caught error. Like `this.setState({ pullingData: false, error })`

src/containers/TokenUpload/index.ts
```diff
@@ -9,32 +14,26 @@ import { State } from 'types'
...
   mapState,
   {
-    getFileContentFromIPFS,
-    openModal,
+    setIPFSFileHash,
     setNewIPFSCustomListAndUpdateBalances,
     setTokenListType,
-    setUploadFileParams,
+    openModal,
   })(IPFSHOC(TokenUpload) as any)
```
Velenir, line 39: `TokenUpload` is wrapped in `IPFSHOC` twice, here and in `TokenUpload` component export

Velenir (approved):

src/containers/TokenUpload/index.ts
Velenir: Still getting timeouts on seemingly valid hashes.

Velenir: We may need to account for `hash+path` in input
like `QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG/readme`, leading to https://gateway.ipfs.io/ipfs/QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG/readme

W3stside: wow good catch wtf was I thinking?

src/actions/tokenList.ts
W3stside: what do u mean? it uses `customTokenList` and calculates `tokenBalances` `async`hronously

src/containers/TokenUpload/index.ts
W3stside: good point about the `path`ing but I am not really sure rn what do think on that

src/actions/tokenList.ts
Velenir: Ignore me, you're right. I was looking at some other action

Velenir (approved):

PR #260, opened 2018-05-23
Feature/new routing

Velenir:
- [x] Incorporates design updates by Michel
- [x] Properly sets up 404 route
- [x] Incorporates disclaimer
- [x] Sets up disclaimer logic with checkboxes, etc.
- [x] Changes first visit mechanism to redirect to disclaimer and back
- [x] Save `disclaimer_accepted` to localStorage


src/components/Disclaimer/index.tsx
```diff
@@ -8,143 +7,188 @@ export interface DisclaimerProps {
...
+      formInvalid: !this.form.checkValidity()
+    })
+  }
+
+  render() {
+    const { formInvalid } = this.state
+
+    let disclaimerConfirmClasses = 'buttonCTA'
+    if (formInvalid) {
+      disclaimerConfirmClasses += ' buttonCTA-disabled'
```
W3stside: iiiinteresting!

src/components/Disclaimer/index.tsx
```diff
@@ -8,143 +7,188 @@ export interface DisclaimerProps {
...
-      <input id="disclaimer5" type="checkbox" required/>
-      <label htmlFor="disclaimer5">
-        <b>
-          The above constitutes a summary of the critical points but is not limited to it.
+            <div className="disclaimerBox md-checkbox">
+              <input id="disclaimer5" type="checkbox" required />
+              <label htmlFor="disclaimer5">
+                <b>
+                  The above constitutes a summary of the critical points but is not limited to it.
           You confirm that you have read the full disclaimer <a href="">available here</a>.
```
W3stside: TODO here so we dont forget

Velenir: Actually that's a `href` to the current page, not a _to-be-added_ href

Velenir: If it for some reason doesn't lead to `./diclaimer` for you, then it's a bug

W3stside (approved):

src/actions/settings.ts
```diff
@@ -0,0 +1,23 @@
...
+  const settings = await localForage.getItem('settings') as Settings
+
+  if (settings) {
+    return dispatch(saveSettings(settings))
+  }
+}
+
+export const asyncSaveSettings = (payload: Partial<Settings>) =>
+  async (dispatch: Dispatch<Settings>, getState: () => State) => {
+    const action = dispatch(saveSettings(payload))
```
W3stside, line 19: why not just return the `dispatch` or `localForage` save directly ? why cache action?

Velenir: Traditionally `dispatch(action)` returns that same action, thunks return `Promise<action>`. By analogy I return the `action`. Doesn't really matter

PR #261, opened 2018-05-28
Feature/dx 311 auction status progress bar

Velenir:
Progressbar in `AuctionStatus` panel depends on current accounts state for the current auction

Progress steps:
  ☐ 1 seller has balance -Deposit Confirmed
  ☐ 2 seller has balance and Auction.ACTIVE - Auction Started
  ☐ 3 auction ended and seller has balance (claimable) - Claim Tokens
  ☐ 4 seller has just claimed - only for visually filling the progress bar
      afterwards on page reload can't determine what has been claimed anyway
  ☐ otherwise progress step = 0

### DIFF
--- a/src/api/dutchx.ts
+++ b/src/api/dutchx.ts
@@ -141,7 +141,7 @@ async function init(): Promise<DutchExchange> {
     index: Index,
     amount: Balance,
     userAccount: Account,
-    ) => dx.claimAndWithdraw(t1, t2, userAccount, index, amount, { from: userAccount })
+    ) => dx.claimAndWithdraw(t1, t2, userAccount, index, amount, { from: userAccount, gas: 4712388 })
 
   const isTokenApproved = (tokenAddress: Account) => dx.approvedTokens(tokenAddress)
 
--- a/src/api/dutchx.ts
+++ b/src/api/dutchx.ts
@@ -86,7 +86,7 @@ async function init(): Promise<DutchExchange> {
     { sell: { address: t1 }, buy: { address: t2 } }: TokenPair,
     index: Index,
     userAccount: Account,
-  ) => dx.claimSellerFunds(t1, t2, userAccount, index, { from: userAccount })
+  ) => (console.log(t1, t2, userAccount, index), dx.claimSellerFunds(t1, t2, userAccount, index, { from: userAccount, gas: 4712388 }))
 
   claimSellerFunds.call = (
     { sell: { address: t1 }, buy: { address: t2 } }: TokenPair,
--- a/src/api/dutchx.ts
+++ b/src/api/dutchx.ts
@@ -86,7 +86,7 @@ async function init(): Promise<DutchExchange> {
     { sell: { address: t1 }, buy: { address: t2 } }: TokenPair,
     index: Index,
     userAccount: Account,
-  ) => (console.log(t1, t2, userAccount, index), dx.claimSellerFunds(t1, t2, userAccount, index, { from: userAccount, gas: 4712388 }))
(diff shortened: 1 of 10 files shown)


src/components/AuctionStateHOC/index.tsx
```diff
@@ -73,6 +74,20 @@ const getAuctionStatus = ({
   return AuctionStatus.INACTIVE
 }
 
+interface ProgressStepArgs {
+  status: AuctionStatus,
+  sellerBalance: BigNumber,
+}
+const getProgressStep = ({ status, sellerBalance }: ProgressStepArgs) => {
+  if (!sellerBalance.gt(0) || status === AuctionStatus.INACTIVE) return 0
```
W3stside: `sellerBalance.lte(0)` no? seems weird with the `!` and `gt` (nitpicking)

src/components/AuctionStateHOC/index.tsx
```diff
@@ -73,6 +74,20 @@ const getAuctionStatus = ({
...
+
+  if (status === AuctionStatus.INIT || status === AuctionStatus.PLANNED) return 1
+
+  if (status === AuctionStatus.ACTIVE) return 2
+
+  if (status === AuctionStatus.ENDED) return 3
+}
```
W3stside: default case? maybe return `INACTIVE` as a failsafe or throw error if no return value

src/components/AuctionStateHOC/index.tsx
```diff
@@ -199,6 +225,16 @@ export default (Component: React.ClassType<any, any, any>): React.ClassType<any,
...
     }
 
+    claimSellerFunds = () => {
+      const { sell, buy, index, account, userCanClaim } = this.state
+      
+      console.log(
+        `claiming tokens for ${account} for
+        ${sell.symbol || sell.name || sell.address}->${buy.symbol || buy.name || buy.address}-${index}`,
+      )
+      return claimSellerFundsAndWitchdraw({ sell, buy }, index, userCanClaim, account)
```
W3stside: should be `claimSellerFundsAndWithdraw` (no 'c')

src/components/AuctionStateHOC/index.tsx
Velenir: Good idea, will default to `0`

src/components/AuctionStateHOC/index.tsx
Velenir: yeah, agree

src/components/AuctionStateHOC/index.tsx
Velenir: No, no, its `witch-draw`, actually but in camelCase

src/api/dutchx.ts
```diff
@@ -86,7 +86,7 @@ async function init(): Promise<DutchExchange> {
     { sell: { address: t1 }, buy: { address: t2 } }: TokenPair,
     index: Index,
     userAccount: Account,
-  ) => dx.claimSellerFunds(t1, t2, userAccount, index, { from: userAccount })
+  ) => dx.claimSellerFunds(t1, t2, userAccount, index, { from: userAccount, gas: 4712388 })
```
W3stside, line 89: how did you get this number? `web3.estimateGas`?

Velenir: You got this number, long time ago, for similar transactions:)

W3stside: should probably check it via `estimateGas` and then `2x` it to be safe

W3stside (approved):
coolio

PR #264, opened 2018-06-02
DX-326-Fix/ipfs default tokens

W3stside:
1. User redirected to TokenOverlay component if CustomList or CustomHash exists in localForagea
2. Default buyToken = undefined
3. Fixed logic to account for this
4. types


src/actions/tokenOverlay.ts
```diff
@@ -20,25 +20,27 @@ export const selectTokenPairAndRatioPair = (props: PropsType) => async (dispatch
...
   try {
     // TODO: dispatch getClosingPrice action
     // which would also add closingPrice to TokenPair state
-    const price = (await closingPrice({ sell, buy })).toString()
+    const promisedPrice = (closingPrice({ sell, buy }))
 
-    dispatch(setClosingPrice({ sell: sell.symbol, buy: buy.symbol, price }))
     dispatch(selectTokenAndCloseOverlay({ mod, token }))
+
+    const price = (await promisedPrice).toString()
```
Velenir: Why start `promisedPrice` above instead of `price = (await closingPrice()).toString()`

Also `promisedPrice` returns `Promise<BigNumber, BigNumber>`, are you sure you want to `[BN, BN].toString()`? That'll be `"BN,BN"`
Why did we even do it like that?!

Velenir (approved):

src/actions/tokenOverlay.ts
W3stside: that is a good point, makes no sense to me looking at it now

PR #265, opened 2018-06-04
Feature/DX-281-OWL-Approval

W3stside:
Changes:

> synchronous biz logic for `checkUserStateAndSell`
1. fires correct modals based on user state
2. passes `accept` from `Promise` as `onClick`
3. uses response to properly fire modals
4. changed `approveAndSell` to `approveTokens` as Sell mechanism removed

> components
1. Extracted hard coded logic into props for `ApprovalModal` + types

### DIFF
--- a/src/actions/blockchain.ts
+++ b/src/actions/blockchain.ts
@@ -207,8 +207,12 @@ export const checkUserStateAndSell = () => async (dispatch: Dispatch<any>, getSt
   } = getState(),
     sellName = sell.symbol.toUpperCase() || sell.name.toUpperCase() || sell.address,
     nativeSellAmt = await toNative(sellAmount, sell.decimals),
+    { TokenOWL } = await promisedContractsMap,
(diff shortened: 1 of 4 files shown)


src/actions/blockchain.ts
```diff
@@ -237,22 +241,53 @@ export const checkUserStateAndSell = () => async (dispatch: Dispatch<any>, getSt
...
+                buttonDesc2: 'Choose this option if you do not want to use OWL',
+              },
+            },
+            onClick: accept,
+          },
+        }))
+      })
+      const choice = await promisedChoice
+
+      await dispatch(approveTokens(choice, 'OWLTOKEN'))
```
Velenir, line 285: Wait, does this mean, you still trigger approval (modal et al) if choice is `Disallow`?

Velenir: No, it doesn't, ignore me

Velenir (approved):

PR #266, opened 2018-06-04
Feature/dx 327 last price

W3stside:
1. Comps lastPrice
2. fixes going forward without choosing tokens

### DIFF
--- a/src/components/TokenPicker/index.tsx
+++ b/src/components/TokenPicker/index.tsx
@@ -12,6 +12,7 @@ export interface TokenPickerProps {
   to: string;
   needsTokens: boolean;
   showPair: boolean;
+  tokensSelected: boolean;
 }
 
 const TokenPicker: React.SFC<TokenPickerProps> = ({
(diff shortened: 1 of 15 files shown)


src/components/TokenPicker/index.tsx
```diff
@@ -32,7 +34,7 @@ const TokenPicker: React.SFC<TokenPickerProps> = ({
       <div className="tokenIntro">
         <h2>Pick Token Pair Auction</h2>
         <TokenPair />
-        <ButtonCTA onClick={continueToOrder} to={to}>Specify amount selling</ButtonCTA>
+        <ButtonCTA className={!tokensSelected ? 'buttonCTA-disabled' : 'blue'} onClick={continueToOrder} to={to}>Specify amount selling</ButtonCTA>
```
Velenir: What about also explicitly blocking button clicks with `disabled={!tokensSelected}`?

Velenir (approved):
Really like simplification without `findRatioPairs`

PR #267, opened 2018-06-06
feature/DX-329-changed claim function for MenuAuction

W3stside:
1. function now withdraws funds
2. fixed error with using wrong indices

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -392,20 +392,30 @@ export const claimSellerFundsFromSeveralAuctions = async (
 
   const claimableIndices = (
     await DutchX.getIndicesWithClaimableTokensForSellers({ sell, buy }, userAccount, indices)
-  )[0].map((i: BigNumber) => i.toNumber())
+  )[0].map(i => i.toNumber())
 
   if (claimableIndices.length === 0) return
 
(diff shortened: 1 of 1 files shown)


src/api/index.ts
```diff
@@ -392,20 +392,30 @@ export const claimSellerFundsFromSeveralAuctions = async (
...
 
+  // getIndicesWithClaimableTokensForSellers returns auctions with sellBalance > 0
+  // this means currentAucIndx may have > 0 sellBalance BUT NOT HAVE CLEARED (only last Index though)
   const lastIndexCleared = (
     await DutchX.getClosingPrice({ sell, buy }, claimableIndices[claimableIndices.length - 1])
   )[0].gt(0)
 
   const length = lastIndexCleared ? claimableIndices.length : claimableIndices.length - 1
+
+  const finalIndices = claimableIndices.slice(0, length)
```
Velenir, line 407: Nice spotting skills! It used to revert tx, right?

Velenir (approved):

src/api/index.ts
W3stside: yep! thanks to @josojo as well as we spotted it together

PR #269, opened 2018-06-06
Feature/dx 331 loaders in modals

W3stside:
> added ETHLogo to Loader and renamed other SVG for use with props  …
1. ETHLogo => tx modals
2. DutchXLogo => other async
3. scss changes for each Modal
4. types

> fixed checkState approval flow + added loader prop  …
1. approveTokens was not throwing error modal and still submittingSellOrder
2. passed loader boolean to TX modals


Velenir (approved):
good work! merge

PR #272, opened 2018-06-06
hotfix/DX-272-Show-Correct-EtherScan-Link

W3stside:
Was always showing `etherscan.io/../../` 

Now shows different link based on `network` from `state`

### DIFF
--- a/src/components/Modals/index.tsx
+++ b/src/components/Modals/index.tsx
@@ -1,6 +1,8 @@
 import React from 'react'
 import { Modal } from 'types'
 import { closeModal } from 'actions'
+import { network2URL } from 'globals'
+
 import Loader from 'components/Loader'
 
 interface TransactionModalProps {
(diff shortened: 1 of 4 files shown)


src/components/Modals/index.tsx
```diff
@@ -47,7 +47,7 @@ export const TransactionModal: React.SFC<TransactionModalProps> = ({
         <li>{`Token Receiving:  ${txData.tokenB.symbol || txData.tokenB.name}${txData.tokenB.name && txData.tokenB.symbol && ' [' + txData.tokenB.name + ']'}`}</li>
         }
         <li>{`Receiving Token Address:  ${txData.tokenB.address}`}</li>
-        <li>Verify receiving token validity via EtherScan: <a target="_blank" href={`https://etherscan.io/token/${txData.tokenB.address}`}>{`https://etherscan.io/token/${txData.tokenB.address}`}</a></li>
+        <li>Verify receiving token validity via EtherScan: <a target="_blank" href={`https://${txData.network === 'RINKEBY' ? 'rinkeby.' : ''}etherscan.io/token/${txData.tokenB.address}`}>{`https://${txData.network === 'RINKEBY' ? 'rinkeby.' : ''}etherscan.io/token/${txData.tokenB.address}`}</a></li>
```
Velenir: this is great, but it only differentiates between `rinkeby` and not `rinkeby`. What about `${network2domain[txData.network] || 'localhost'}`?

Velenir (approved):
I like small PRs

src/components/Modals/index.tsx
W3stside: i can dig that

src/globals/index.ts
```diff
@@ -12,6 +12,13 @@ export const code2tokenMap: Code2Name = {
   REP: 'AUGUR',
 }
 
+export const network2URL: Network2URL = {
+  RINKEBY: 'https://rinkeby.etherscan.io/',
+  MAIN: 'https://etherscan.io/',
+  KOVAN: 'https://kovan.etherscan.io/',
+  UNKNOWN: 'localhost:5000/',
```
Velenir: So this will create `localhost:5000/token/address` which leads nowhere? I guess it doesn't really matter
Also do `//localhost:500/` to match protocol on localhost only

W3stside: ya

PR #277, opened 2018-06-15
Feature/dx 343 fix default list

W3stside:
1. `blockchain > calcAllTokenBalances` fix calcing shit for no reason
2. set conditional NODE_ENV pulling of tokens in `walletintegrations`
3. `ETHER` added as token checked against users account (change this to random key we can just pass in `api/index`?? (instead of async fn call)
4. `toFixed(2) > toFixed(4)` in `MenuWallet`


W3stside:
@Velenir changing to re-addd in `TokenMGN` and re-import in `api/contracts` to show correct `feeRatio` at top - remove from displaying in `MenuWallet`

Velenir (approved):

PR #281, opened 2018-06-20
Fix/order refresh - redirect to home

Velenir:
Redirects home if Metamask isn't unlocked

- if `blockchain.currentAccount === undefined` redirects to /
- moved `initializer()` from `React.render` callback to before `render` to fill in state beforehand


W3stside (approved):
cool, merge

PR #288, opened 2018-06-27
Feature/dx 330 adding content pages

W3stside:
Sorry this PR seems so big - it's a lot of removals of dead code and adding static content pages

Just focus on the `claimable` prop being passed in `MenuAuctions` and the selector in `selector/auctions`


src/components/Hamburger/index.tsx
```diff
@@ -21,18 +22,25 @@ export default class Hamburger extends Component<HamburgerProps, HamburgerState>
...
+        <button className="hamburger" onClick={this.handleClick}></button>
         <nav className={isOpen ? 'show' : null}>
           <button
             className="buttonExit"
             onClick={this.handleClick}>
           </button>
-          <a href="#">How it works</a>
-          <a href="#">About</a>
-          <a href="#">Faq</a>
+          <Link replace to="/content/HowItWorks"> How It Works </Link>
```
Velenir: Why replace? It makes me unable to go back to `/` easily

src/router.tsx
```diff
@@ -1,40 +1,42 @@
...
   disabled?: boolean;
 }
 
 // TODO: consider redirecting from inside /order, /wallet, /auction/:nonexistent_addr to root
-
-const withHeader = (Component: React.ComponentClass) => (props: any) => (
+const withHeaderAndFooter = (Component: React.ComponentClass | React.SFC, content?: boolean) => (props: any) => (
   <Fragment>
-    <Header/>
+    <Header content={content}/>
```
Velenir: What about instead of a case-specific `content` we pass in `{ headerProps, footerProps }`,  and then destructure them onto `Header, Footer` respectively. Doesn't ultimately matter, of course.

src/selectors/auctions.ts
```diff
@@ -0,0 +1,7 @@
+import { createSelector } from 'reselect'
+import { State } from 'types'
+
+export const auctionClaimable = createSelector(
+  (state: State) => state.auctions.ongoingAuctions,
+  auctions => auctions.find((auc: any) => auc.claim || auc.claimInverse),
```
Velenir: You a passing in the whole auction instead of a bool. Use `auctions.some(...)`.

src/components/Hamburger/index.tsx
```diff
@@ -21,18 +22,25 @@ export default class Hamburger extends Component<HamburgerProps, HamburgerState>
...
+          <Link replace to="/content/Screencast">Screencast</Link>
+          <Link replace to="/content/Tokens">Tokens </Link>
+          <Link replace to="/content/Fees">Fees</Link>
+          <Link replace to="/content/FAQ">FAQ</Link>
+          <Link replace to="/content/Technical">Technical </Link>
+          <Link replace to="/content/Downtime">Downtime</Link>
+          <Link replace to="/content/Help">Help</Link>
```
Velenir: You may want to add `onClick={this.handleClick}` to all Links to close the menu. Or not. Depending on the intended UX.

src/components/ContentPage/index.tsx
```diff
@@ -0,0 +1,636 @@
...
+      <p>Auctions end when the auction clearing price is reached and all sellTokens are being bought and thus allocated to bidders at this final price (i.e. bidVolume x price = sellVolume).</p>
+    </section>
+    <section className="drawer">
+      <h3>Which tokens can I exchange?</h3>
+      <p>You may find the tokens available to trade on this interface within the token list (click on either the sell or receive token). Note that all tokens are always available to trade with ETH (or WETH) but may not yet exists as a pairing with one another.</p>
+    </section>
+    <section className="drawer">
+      <h3>What do I need to exchange a token on the DutchX?</h3>
+      <p>You only need three things:
+      <ol>
```
Velenir: Generates `Warning: validateDOMNesting(...): <ol> cannot appear as a descendant of <p>.`. aka _Inside `<p> `phrasing content only, which `<ol>` isn't_ 
A LOT. Console is filled with them. Also in FAQ and Fees

src/components/Header/index.tsx
```diff
@@ -6,8 +6,12 @@ import Hamburger from 'components/Hamburger'
...
 import { Link } from 'react-router-dom'
 
-export const Header: React.SFC = () => (
-  <header>
+interface HeaderProps {
+  content?: boolean;
+}
+
+export const Header = ({ content }: HeaderProps) => (
+  <header className={content && 'solid-background'}>
```
Velenir: Generates `warning.js?1792:33 Warning: Received false for a non-boolean attribute className.` Do a ternary with empty string or use my suggestion about passing `headerProps`

src/styles/components/navbar/_navbar.scss
```diff
@@ -15,24 +15,44 @@ header {
```
Velenir, line 15: Can you remove `header {width: 100vw }` a few lines above. It adds a horizontal scrollbar when a vertical appears at some sizes.

Velenir (changes requested):
With the addition of the `<Footer>` some pages suffer from the _css holy grail_ problem.
Footer appears as soon as it can fit after the main content, which can be easily seen on vertically short pages like `/content/Help`

There are many ways to fix it, many depend on page structure, easiest that comes to mind:

```
#root > div {
    min-height: 100vh;
    display: flex;
    flex-direction: column;
}

footer { margin-top: auto; }
```

src/components/Hamburger/index.tsx
W3stside: i was testing replace and never changed it, thanks

src/router.tsx
W3stside: /shrug

src/selectors/auctions.ts
W3stside: ah, very cool!

src/components/ContentPage/index.tsx
W3stside: mm yeah HTML errors, yay

src/styles/components/navbar/_navbar.scss
W3stside: done

Velenir (approved):
Titanic work! approved

PR #294, opened 2018-07-05
Fix planned auctions not included in Your Auctions

Velenir:
Also check for sellVolumeNext when `indicesWithSellerBalance` is empty

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -626,11 +626,20 @@ export const getSellerOngoingAuctions = async (
       let ongoingAuction: AuctionObject
 
       const { sell: { decimals }, buy: { decimals: decimalsInverse } } = auction
-      if (indicesWithSellerBalance.length >= 1 || indicesWithSellerBalanceInverse.length >= 1) {
+      if (
+        indicesWithSellerBalance.length >= 1 || indicesWithSellerBalanceInverse.length >= 1 ||
+        // also check for no claiming indices, but potentially sellVolumeNext
+        indicesWithSellerBalance.length === 0 || indicesWithSellerBalanceInverse.length === 0
+      ) {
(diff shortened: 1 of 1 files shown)


src/api/index.ts
```diff
@@ -626,12 +626,21 @@ export const getSellerOngoingAuctions = async (
       let ongoingAuction: AuctionObject
 
       const { sell: { decimals }, buy: { decimals: decimalsInverse } } = auction
-      if (indicesWithSellerBalance.length >= 1 || indicesWithSellerBalanceInverse.length >= 1) {
+      if (
+        indicesWithSellerBalance.length >= 1 || indicesWithSellerBalanceInverse.length >= 1 ||
```
W3stside: just use `>=` 0 here

Velenir: should there even be `if`

W3stside: haha

src/api/index.ts
```diff
@@ -625,56 +625,41 @@ export const getSellerOngoingAuctions = async (
...
-        const committedToNextNormal = sellVolumeNext.normal.gt(0)
-        const committedToNextInverse = sellVolumeNext.inverse.gt(0)
-
-        if (  // if there are truly no auctions with sellBalance
-          !committedToNextNormal && !committedToNextInverse &&
-          indicesWithSellerBalance.length === 0 && indicesWithSellerBalanceInverse.length === 0
-        ) return accum
-
-        let latestIndicesNormal: BigNumber[] = indicesWithSellerBalance,
+      const { sell: { decimals }, buy: { decimals: decimalsInverse } } = auction\
```
W3stside: / here

W3stside (approved):

PR #295, opened 2018-07-05
Auction status revamped

Velenir:
Changes in AuctionPanel and its children


W3stside:
@Velenir submitted suggested changes via #296

W3stside (approved):

PR #296, opened 2018-07-05
Feature/auction status revamped david edits

W3stside:
1. Design changes (CSS + text)
> If not claimable design changes
> added `<h3>` tag for displaying messages

2. Do not render bottom blue `AuctionFooter` if `Status === Ended` + claimed

### DIFF
--- a/src/components/AuctionFooter/index.tsx
+++ b/src/components/AuctionFooter/index.tsx
@@ -24,7 +24,7 @@ const AuctionFooter: React.SFC<AuctionFooterProps> = ({
   // buyDecimal,
   status,
 }) => {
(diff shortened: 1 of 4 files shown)


src/components/AuctionFooter/index.tsx
```diff
@@ -58,9 +58,7 @@ const AuctionFooter: React.SFC<AuctionFooterProps> = ({
   }
 
 
-  return (
-    <div className="auctionFooter"/>
-  )
+  return null
```
Velenir: Consider a gradual footer removal rather than it just disappearing from under you.

Velenir (approved):

PR #298, opened 2018-07-09
Fix/wallet page flash

Velenir:
When postSellOrder modal is closing we don't immediately go to the `auction/*` path, only after refetching balance. So `/wallet` panel is shown for a second.
This goes to the `auction/*` path right away.


W3stside (approved):
nice, tiny PRs are the tits

PR #306, opened 2018-07-14
Feature/dx balances menu wallet

W3stside:
# Adds DX Token Balances to `MenuWallet`


W3stside:
closed as already inside #319

PR #314, opened 2018-07-30
Fix/DX-452_DX-466_DX-472_bug-fixes

W3stside:
### Bug Fixes

Fixes:
1. New Token Lists not updating automatically
2. Incorrect redirection links in Content pages
3. MGN bug (466)


Velenir (approved):
Looks great, especially like going back to jsons for tokenList, as it meant to be.

PR #323, opened 2018-08-03
Fix DX-477

Velenir:
Fixed height didn't allow `section.home` to expand together with its children.


W3stside:
why `develop` merge again?

W3stside (approved):
Looks good!

PR #325, opened 2018-08-07
Dx 501 auction times

Velenir:
Adds time section to running and planned auctions
WIP, pending likely changes to the ticket.


W3stside (approved):
Just make that p margin top edit and the break tag :)

PR #332, opened 2018-08-16
Dx 486/feature/your auctions redux

Velenir:
Displaying current/past/next auctions per direction per pair in YourAuctions.
With claim time, claim amount and status.


W3stside (approved):
Approving and if errors come up, we can address them, nice work!

PR #333, opened 2018-08-17
check availability as soon as registering provider

Velenir:
Check for availability of injected provider as soon as.


W3stside:
this will have to be refactored w/Ledger integration but approving for now!

W3stside (approved):

PR #337, opened 2018-08-24
DX-540-DX-531/Cookie-Monster-Footer-Fantasia

W3stside:
Cookie component + saving to localForage

WIP - needs Footer links, etc


src/components/Cookies/index.jsx
```diff
@@ -0,0 +1,241 @@
...
+        necessary: true,
+        analytics: false,
+      },
+      loading: false,
+      error: undefined,
+    }
+
+    async componentDidMount() {
+      try {
+        this.setState({ loading: true })
```
Velenir: why `loading: true` here as opposed to in constructor?

W3stside: :+1:

PR #338, opened 2018-08-28
Dx 548/update geo block

W3stside:
Incorporates logic required for release as seen here:
```
Make sure that:
In rinkeby version: Only rinkeby network is permitted
In mainnet version: Only Mainnet network is permitted
...
&&
...
3. No Geoblock at all
...
```

@Velenir please check my `conditionalRender` logic in `src/index.tsx`

Also - what branch are we now merging into? Happy to open a new `release/v1.0.2` branch or whatever it needs to be called


src/index.tsx
```diff
@@ -32,23 +32,51 @@ const preAppRender = async () => {
...
+    if (blocked) {
+      disabledReason = 'networkblock'
+
+      window.history.replaceState(null, '', '/')
+      return rootElement.innerHTML = ReactDOMServer.renderToStaticMarkup(<App disabled disabledReason={disabledReason} networkAllowed={ALLOWED_NETWORK}/>)
+    }
   }
+  else if (hostname === URLS.DUTCHX_APP_URL_MAIN) {
+    ALLOWED_NETWORK = 'Ethereum Mainnet'
+    const netBlockedPromise = isNetBlocked('1')
```
Velenir: So we block rinkeby, kovan, testrpc on `dutchx.app`, correct?

src/block.ts
```diff
@@ -4,44 +4,38 @@ export const geoBlockedCountryCodes = new Set(blocked_codes)
...
 
-export const isNetBlocked = async () => {
-  if (typeof window === 'undefined') return false
-  const { hostname } = window.location
-    // allow anything when run locally
-  if (hostname === 'localhost' || hostname === '0.0.0.0') return false
-
-    // no walletextension detected, different error - download wallet error
-  if (!window.web3) return false
+export const isNetBlocked = async (idToAllow: string | number) => {
```
Velenir: You sure we ever allow only one network? I'm thinking maybe use a networkId array just in case

Velenir:
Let's do a `release/` branch

Velenir (approved):

PR #343, opened 2018-09-05
DX-547/React-Google-Analytics

W3stside:
Adds GA to dx-react

There is an annoying issue where testing React.GA is not initialised thus it logs bs to the console

Haven't tested it accurately tracks page views but looks to be no other problems otherwise - waiting for GA login info


src/components/WithTrackingHOC/index.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { Component } from 'react'
+import ReactGA from 'react-ga'
+
+interface WithTrackerHOCProps extends Window {}
+
+export default function withTracker(WrappedComponent: React.SFC | React.ComponentClass, options = {}) {
```
Velenir: No need to wrap every component on every route. This can be just one component rendered on every route, it only needs to have access to Router props.
Like:
```
export default withRouter(class extends Component<WithTrackerHOCProps> {
...
```

And put it below `<RedirectToDisclaimer/>`

src/components/WithTrackingHOC/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+    componentDidMount() {
+      const page = this.props.location.pathname + this.props.location.hash
+      trackPage(page)
+    }
+
+    componentWillReceiveProps(nextProps: any) {
+      const currentPage = this.props.location.pathname + this.props.location.hash
+      const nextPage = nextProps.location.pathname + nextProps.location.hash
+
+      if (currentPage !== nextPage) {
```
Velenir: This would trigger when user goes between different hashes inside `/content/Tokens` path.
React-router already correctly parses the url like `/#/content/Tokens#what-is-mgn` into pathname `"/content/Tokens"` and hash thanks to HashHistory

src/components/WithTrackingHOC/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+        trackPage(nextPage)
+      }
+    }
+
+    render() {
+      return <WrappedComponent {...this.props} />
```
Velenir: If you do a non-wrapper like I suggest, there'll be no need for `render()` at all.

Velenir (dismissed):

src/components/WithTrackingHOC/index.tsx
W3stside: not sure what to do here then

src/components/WithTrackingHOC/index.tsx
W3stside: is that better? more efficient or just DRY? Just trying to undershtand

Velenir: all of them. With HOC you'll have 
```
<Wrapper>
 <Component>
</Wrapper>
```
two components for one functionality with parent retriggering child render every time

With one component always there in the tree and if it also doesn't render anything, impact is minimal

src/components/WithTrackingHOC/index.tsx
Velenir:
```
      const currentPage = this.props.location.pathname
      const nextPage = nextProps.location.pathname
````

W3stside: pathname will always be `/` with our setup i believe

W3stside: fixed

PR #344, opened 2018-09-05
DX-446: Home screen wording

W3stside:
Changes `<Home>` component wording as per `DX-446`

### DIFF
--- a/src/components/TextSquare/index.tsx
+++ b/src/components/TextSquare/index.tsx
@@ -1,38 +1,34 @@
 import React from 'react'
 import ButtonCTA from 'components/ButtonCTA'
 
-export const TextSquare: React.SFC = () => {
-  return (
-    <div className="intro">
-      <h1>Decentralized Token Auction Exchange</h1>
-      <div className="textSquareInnerContainer">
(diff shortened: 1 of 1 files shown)


src/components/TextSquare/index.tsx
```diff
@@ -1,38 +1,34 @@
...
+        <li>Submit the order via your wallet provider</li>
+      </ol>
+
+      <br />
+
+      Your order is automatically deposited into the next running auction - no strategy needed!
+
+      <br /><br />
+
+      <ButtonCTA className="void" to="/content/HowItWorks">How it works</ButtonCTA>
```
Velenir: Pass `null` in `className`, `"void"` gives it `.void` class unnecessarily.
Why did I notice it just now?  :unamused:

Velenir (approved):

src/components/TextSquare/index.tsx
W3stside: oh woops i dont even know how i kept that in there i actually had a class named vvoid at some point  i guess ahah. thanks

W3stside: fixed in #343

PR #346, opened 2018-09-07
DX-557/timing wording change

Velenir:
`6:45h` -> `6h:45min`

localhost_5000_ 2


W3stside:
small annoying ask - is it possible to change `13h` to `13 hours`? 

`13h` looks too much like `13:00` - in french we state time via `12h` or `17h` to say it is 12 o clock or 17 o clock

W3stside (approved):

PR #347, opened 2018-09-07
Dx-486/rebased backport

Velenir:
Here lie the YourAuctions' timings.
I would really like to merge this one and get it into production
Otherwise while all the other stuff gets written and this feature rests in alt-refactor conflicts only multiply


W3stside (approved):
works and looks good to me

PR #348, opened 2018-09-07
DX-525/ATR

Velenir:
Template for Add Token Requests


W3stside (approved):
approving as looks fine. can be re-iterated later

PR #352, opened 2018-09-11
Release/v1.0.3

W3stside:
`Provider` logic changes

Open to suggestions on what the Object looks like


webpack.dev.config.js
```diff
@@ -125,6 +125,7 @@ module.exports = {
   devServer: {
     disableHostCheck: true,
     historyApiFallback: true,
+    https: false,
```
Velenir: It's `false` by default. Was there a problem before?

Velenir (approved):

src/integrations/provider/index.ts
```diff
@@ -1,29 +1,138 @@
...
+// const rpcUrl = network2RPCURL.RINKEBY
+// const networkId = 4 // parseInt(process.env.REACT_APP_NETWORK_ID || "1337", 10);
+// ====================================================================================
+
+const Providers = {
+  // runtime providers (METAMASK/MIST/PARITY)
+  INJECTED_WALLET: {
+    priority: 90,
+    providerType: 'INJECTED_WALLET',
+    keyName: 'INJECTED_WALLET',
```
Velenir, line 64: What's `keyName` about?

PR #364, opened 2018-09-14
Fix Loader flashing when being redirected to auction

Velenir:
fix the whole Approuter unmounting


W3stside (approved):
such a simple change with big implications, nice!

PR #365, opened 2018-09-17
Dx 446-Release-1.0.4/Terms-And-Conditions

W3stside:
Adds T & C's

1. adds new routes required to be accessible BEFORE app load inside disclaimer:
  a. `Cookies`
  b. `HowItWorks`
  c. `Privacy`
2. T & Cs route added - Privacy re-routes to home until its available
3. Selector conditional checks if `provider` / `activeProvider` not yet avail - hacky workaround logic dont like it
4. `WalletIntegration` + `AppValidator` use validation prop (disclaimer_accepted) to skip mount logic if met, also dont like it but prevents app from loading all data etc if user checks out internal routes from `Disclaimer` before accepting


src/utils/index.ts
```diff
@@ -178,3 +178,25 @@ export const provider2SVG = (providerName: ProviderName | ProviderType) => {
...
   }
 }
+
+export const web3CompatibleNetwork = () => {
+  if (typeof window === 'undefined' || !window.web3) return 'UNKNOWN'
+
+  let netID
+
+  // 1.X.X API
+  if (typeof window.web3.version === 'string') {
```
Velenir, line 188: Should this be `typeof window.web3.version !== 'string`?

src/utils/index.ts
```diff
@@ -178,3 +178,25 @@ export const provider2SVG = (providerName: ProviderName | ProviderType) => {
...
+    window.web3.eth.net.getId((err: Error, res: string) => {
+      if (err) {
```
Velenir: This has to be a Promise, cause callback.

src/components/Footer/index.tsx
```diff
@@ -25,7 +25,7 @@ const Footer = ({ network }: FooterProps) =>
                 </>
                     :
                 <>
-                    <Link to="/privacy">Privacy Policy</Link><Link to="/cookies">Cookies</Link><Link to="/terms">Terms of Use</Link><Link to="/imprint">Imprint</Link> <span> Protocol built by <img src={gnosisLogoSVG} className="footerLogo"/></span>
+                    <Link to="/privacy">Privacy Policy</Link><Link to="/cookies">Cookies</Link><Link to="terms">Terms and Conditions</Link><Link to="/imprint">Imprint</Link> <span> Protocol built by <img src={gnosisLogoSVG} className="footerLogo"/></span>
```
Velenir: `/terms` ?

Velenir (changes requested):

src/utils/index.ts
W3stside: nay

src/components/Footer/index.tsx
W3stside: yes, good catch!

src/utils/index.ts
W3stside: _cries_

Velenir (approved):

PR #369, opened 2018-09-21
Fixes/mobile bugs hardcode bugs

W3stside:
Fixes bugs:

Mobile: `MenuAuctions` absolute positioning + `left: 0` with new container was hiding container outside of screen. 

Dynamic text to replace hardcoded `MetaMask` provider name in `WalletPanel` screen (just before blockchain modals)

Attempts to fix `noScroll` and scroll behaviour on modals especially on `Final COnfirmation` screen that has a lot of content on mobile. Previous changes was showing non-scrollable Y scrollbar due to `overflow-y: scroll` on at all times

and other stuff....


src/components/App/index.tsx
```diff
@@ -38,4 +38,9 @@ const App = (props: AppProps): any => {
     </Provider>
   )}
 
+// history listen on change
+history.listen((loc: any) => {
+  loc.search.includes('retrox') && document.body.classList.add('RETROX')
```
Velenir: I though about it and a better, more compatible way would be
```js
const searchParams = new URLSearchParams(location.search)
document.body.classList.toggle('RETOX', searchParams.has('retox'))
```

src/components/Hamburger/index.tsx
```diff
@@ -37,7 +37,7 @@ export default class Hamburger extends Component<HamburgerProps, HamburgerState>
             className="buttonExit"
             onClick={this.handleClick}>
           </button>
-          {window && window.location.pathname !== '/' && <Link to="/" onClick={this.handleClick}> Home </Link>}
+          {window && window.location.hash !== '#/' && <Link to="/" onClick={this.handleClick}> Home </Link>}
```
Velenir: Would be better not to depend on window here, as that stuff will change based on Hash/Browser history every time
Try
```
<Route exact path="/" render={() => <Link to="/" onClick={this.handleClick}> Home </Link>}/>
```

src/middlewares/NoScroll.ts
```diff
@@ -1,17 +1,16 @@
...
 const NoScroll = () => (next: Function) => async (action: any) => {
   const { type } = action as Action
-
-  if (action.payload && action.payload.modalProps && action.payload.modalProps.txData || type !== 'CLOSE_MODAL' && type !== 'OPEN_MODAL') return next(action as Action)
+  if (action.payload && action.payload.modalProps && action.payload.modalProps.header !== 'Order confirmation' || type !== 'CLOSE_MODAL' && type !== 'OPEN_MODAL') return next(action as Action)
 
   try {
     if (type === 'OPEN_MODAL') {
       document && document.body && document.body.classList.add('noScroll')
+      document && document.getElementById('root') && document.getElementById('root').classList.add('noScroll')
```
Velenir: Don't search for element twice unnecessarily
```
const root = document && document.getElementById('root')
root && root.classList.add('noScroll')
```

src/middlewares/NoScroll.ts
```diff
@@ -1,17 +1,16 @@
...
     } else if (type === 'CLOSE_MODAL') {
       document && document.body && document.body.classList.remove('noScroll')
+      document && document.getElementById('root') && document.getElementById('root').classList.remove('noScroll')
```
Velenir: same as above

Velenir (commented):
Wondering if we can cut down on it by using css vars.
Let's speak Monday

src/components/Hamburger/index.tsx
W3stside: but this doesn't make sense, render Home when on "/" aka home path? should be the inverse

src/components/App/index.tsx
W3stside: works well!

src/components/Hamburger/index.tsx
Velenir: Yes, a million times
```
<Route path="/(.+)" render={() => <Link to="/" onClick={this.handleClick}> Home </Link>}/>
```
this works

W3stside: cool, good stuff ;)

Velenir (approved):

PR #370, opened 2018-09-24
Dx 572/fix overlapping on mobile

Velenir:
Fixes `.offlineBanner` overlapping everything else on mobile


W3stside (approved):
nice was trying to fix this too

PR #375, opened 2018-09-26
Detailed Account + Wallet view && Change wallet

W3stside:
1. adds detailed Account view in MenuWallet
2. adds ability to change wallet on click


Velenir (approved):
Looks good! Can add Status and co. icons in the future

PR #384, opened 2018-10-02
Feature/adding priv key wallet

W3stside:
## Private Key Wallet round 1... Fight!

1. Allows user to enter Private Key and get a wallet created - right now it only works for `localhost` as it is hard-coded in - TODO: allow user to enter URL of infura node to connect to?
2. changes API as singleton logic was preventing changing wallets (it would check oh is var filled? yes? ok fuck u then no change) - so now this is overridden by a `boolean` parameter on setup. This does present problems as changing wallets and clicking the same will always re-instantiate things. But this works and prevents reloading of API during setup.
3. Creation of crude `validateTransaction` modal like MM does before signing. It sucks because, well it sucks, and mainly because it covers the informative modals telling users what's going on. Need to find a better solution.


ghost:
Travis automatic deployment:
 https://pr384--dxreact.review.gnosisdev.com

ghost:
Travis automatic deployment:
 https://pr384--dxreact.review.gnosisdev.com

ghost:
Travis automatic deployment:
 https://pr384--dxreact.review.gnosisdev.com

ghost:
Travis automatic deployment:
 https://pr384--dxreact.review.gnosisdev.com

PR #389, opened 2018-10-08
Feature/DX-605-Estimate-Gas

W3stside:
Ready
`utils/index.ts` is where the `estimateGas` fn lives - still WIP as edge cases not tested

### DIFF
--- a/src/api/dutchx.ts
+++ b/src/api/dutchx.ts
@@ -1,7 +1,7 @@
 import { promisedContractsMap } from './contracts'
 import { DutchExchange, Index, Filter, ErrorFirstCallback, DutchExchangeEvents } from './types'
 import { TokenPair, Account, Balance, TokenCode } from 'types'
-import { GAS_PRICE, GAS_LIMIT_TESTING } from 'globals'
+import { estimateGas } from 'utils'
 
 let dutchXAPI: DutchExchange
 
@@ -70,7 +70,7 @@ async function init(): Promise<DutchExchange> {
     amount: Balance,
     index: Index,
     userAccount: Account,
-  ) => dx.postSellOrder(t1, t2, index, amount, { from: userAccount, gasPrice: GAS_PRICE, gas: GAS_LIMIT_TESTING })
+  ) => estimateGas({ cb: dx.postSellOrder, mainParams: [t1, t2, index, amount], txParams: { from: userAccount } })
 
   postSellOrder.call = (
     { sell: { address: t1 }, buy: { address: t2 } }: TokenPair,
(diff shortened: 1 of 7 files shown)


src/api/dutchx.ts
```diff
@@ -175,7 +176,7 @@ async function init(): Promise<DutchExchange> {
     { sell: { address: t1 }, buy: { address: t2 } }: TokenPair,
     amount: Balance,
     userAccount: Account,
-  ) => dx.depositAndSell.sendTransaction(t1, t2, amount, { from: userAccount, gasPrice: GAS_PRICE, gas: GAS_LIMIT_TESTING })
+  ) => estimateGas({ cb: dx.depositAndSell, mainParams: [t1, t2, amount], txParams: { from: userAccount } }, 'sendTransaction')
```
anxolin, line 189: Is that `dx.depositAndSell` estimation work? In the past it wasn't working very good the estimate using a proxy

src/utils/index.ts
```diff
@@ -215,3 +215,26 @@ export const web3CompatibleNetwork = async () => {
 }
 
 export const lastArrVal = (arr: Array<any>) => arr[arr.length - 1]
+
+export const estimateGas = async (
+  { cb, mainParams, txParams }: { cb: Function & { estimateGas: Function }, mainParams: any, txParams: TransactionObject },
+  type: null | 'sendTransaction' | 'call',
+) => {
+  let estimatedGasPrice: string
+  const estimatedGasLimit: string | number = await cb.estimateGas(...mainParams, { ...txParams }).catch(() => '4000000')
```
anxolin: What is the 4000000?

ghost:
Travis automatic deployment:
 https://pr389--dxreact.review.gnosisdev.com

ghost:
Travis automatic deployment:
 https://pr389--dxreact.review.gnosisdev.com

src/types/global.d.ts
```diff
@@ -32,3 +32,7 @@ declare const after: typeof afterAll
 declare interface Array<T> {
   __last: () => T;
 }
+
+declare interface Function {
+  estimateGas(mainParams?: any, txParams?: any): any;
```
Velenir: Ugh, every function will have it like that.
Instead in api/types do
```
type DutchExchangeExtended = {
  [K in keyof DutchExchange]: DutchExchange[K] extends (...args: any[]) => Promise<Receipt> ?
    DutchExchange[K] & {sendTransaction?: DutchExchange<Hash>[K], estimateGas?: DutchExchange<Hash>[K]} :
    DutchExchange[K]
}
```
wherever sendTransaction appears add estimateGas

src/utils/index.ts
```diff
@@ -215,3 +215,24 @@ export const web3CompatibleNetwork = async () => {
...
+  } catch (error) {
+    console.warn('Safe gas estimation error: ', error, 'Defaulting to lowest gas price')
+    estimatedGasPrice = '1000000000'
+  }
+
+  console.warn('ESTIMATE GAS FINAL FUNCTION PARAMS: ', mainParams, { ...txParams, gas: estimatedGasLimit, gasPrice: estimatedGasPrice })
+
+  if (!type) return cb(...mainParams, { ...txParams, gas: estimatedGasLimit, gasPrice: estimatedGasPrice })
+
+  return cb[type](...mainParams, { ...txParams, gas: estimatedGasLimit, gasPrice: estimatedGasPrice })
```
Velenir: less duplication
```
(type ? cb[type] : cb)(...mainParams, ...)

**Review by Velenir — CHANGES_REQUESTED** [review]

File: src/types/global.d.ts (hunk 3 above)
Comment (W3stside) [reply src/types/global.d.ts]: ah yes that's what I was looking for!

File: src/utils/index.ts (hunk 4 above)
Comment (W3stside) [reply src/utils/index.ts]: :+1:

**Review by Velenir — APPROVED** [review]

File: src/api/dutchx.ts [hunk 5]
```diff
@@ -70,7 +70,7 @@ async function init(): Promise<DutchExchange> {
     amount: Balance,
     index: Index,
     userAccount: Account,
-  ) => dx.postSellOrder(t1, t2, index, amount, { from: userAccount, gasPrice: GAS_PRICE, gas: GAS_LIMIT_TESTING })
```
anxolin, line 73: This review is not for the current version, and it's just my opinion.
I find easier to use the old approach were u had a constant, but instead of using a single one (`GAS_LIMIT_TESTING`) we could define the gas limit per operation so it's more accurate.

Example in this case, could be sth like:
```js
 dx.postSellOrder(t1, t2, index, amount, { from: userAccount, gasPrice: GAS_PRICE, gas: GAS_LIMITS.postSellOrder }
```

PR #436, opened 2018-12-05
add ID check param to web3Network check to fix block

W3stside:
HOTFIX

1. add ID check param to web3Network check to fix block
  a. due to a change in params in web3CompatibleNetwork fn, block function was checking against network STRING NAME e.g Rinkeby or Main instead of it's ID e.g 1 || 4 || 42 etc


ghost:
Travis automatic deployment:
 https://pr436--dxreact.review.gnosisdev.com

ghost:
Travis automatic deployment:
 https://pr436--dxreact.review.gnosisdev.com

